PR #83, opened 2019-10-17
filter against the many to many field barrier_types

markhigham:
barrier_type looks like it is no longer used.
perhaps create a PR with that field removed


markhigham:
@sekharpanja I've updated the unit tests and removed the old field.

How come this was still in the source?

api/barriers/models.py
```diff
@@ -126,13 +126,7 @@ class BarrierInstance(BaseModel, ArchivableModel):
...
     )
-    barrier_type = models.ForeignKey(
-        BarrierType,
-        null=True,
-        default=None,
-        related_name="barrier_type",
-        on_delete=models.SET_NULL,
-    )
+    
     barrier_type_category = models.CharField(
```
sekharpanja: You can remove this as well. Both barrier_type and category used to be internal fields of this model before they are turned into a many-to-many. They were left for migration purpose and should now be removed.

PR #84, opened 2019-11-04
Add management command for archiving barriers

rossmiller:
Interim solution for archiving barriers.


api/core/management/commands/archive_barrier.py
```diff
@@ -0,0 +1,71 @@
...
+            help="GUID of barrier to archive",
+            dest="barrier_id",
+        )
+
+    def handle(self, *args, **options):
+        username = options["username"]
+        barrier_id = options["barrier_id"]
+
+        if username is None:
+            raise CommandError(
```
markhigham: I think a better error message would be 
```
Please supply the email of the requested user.

Usage: ./manage.py --username username@domain.com --barrier_id guid-for-the-barrier
```

rossmiller: Good catch on the location of the command, I've just moved it to the barriers app. 

I disagree on the error message - the user is the "requester" because they are the person asking for the archive to take place, so "requested" is confusing. Additionally, the reason I put "using --username" is because it is potentially confusing to request an email address using the username argument and I want the command user to know that the username column uses an email address. This message makes that obvious.

The "Usage" component of the message is unnecessary as you can get a list of available arguments using the management command help argument.

PR #85, opened 2019-12-03
MAR-91

fericsepi:
- docker image update, switched to centos 7 with SSH baked in for local development
 - separating requirements to dev and prod
 - introducing Makefile
 - pre commit black formatting
 - switch to use postgres locally


Makefile
```diff
@@ -0,0 +1,80 @@
+# ==================================================
```
rossmiller: The Django shortcuts are nice!

docker-compose.yml
```diff
@@ -1,34 +1,43 @@
...
+      - "8880:8000"
+      - "8882:22"
     depends_on:
-      - web
-    links:
-      - web
-    command: celery worker -A config.celery -l info -B
+      - db
+
+#  celery:
```
rossmiller: When I talked about why Celery was used I was told by @sekharpanja it needed for the anti-virus check. I spoke to @munhitsu about this and he told me that it's not required but it might make sense if large files are being processed.

docker/local/Dockerfile
```diff
@@ -0,0 +1,61 @@
+FROM centos:7
```
rossmiller: Interesting approach! What is the reason for not using a standard Python image?

readme.md
```diff
@@ -2,10 +2,33 @@
...
+    cd market-access-api
+    ```
+4. Copy the env file - `cp docker-compose.env.sample docker-compose.env`         
+
+#### Install
+1. Build the images and spin up the containers by running - `docker-compose up --build`
+2. Set up git hooks by running - `make git-hooks`
+3. To start the dev server run - `make django-run`
+4. The API is now accessible via http://api.market-access.local:8880
+
```
rossmiller: Just followed the instructions and got the following error from the web container :

```django.core.exceptions.ImproperlyConfigured: Set the RESOURCE_SERVER_USER_INFO_URL environment variable```

File: tools/git_hooks/pre-commit.sh [hunk 5]
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
rossmiller: Have not tried this but looks like a cool idea!

docker/local/Dockerfile
fericsepi: I used this image before and it has SSH in here :)

readme.md
fericsepi: thanks, yes I need to provide a more sensible env sample file - commit inc

docker-compose.yml
fericsepi: right thanks, I'll get celery wrapped up too

readme.md
fericsepi: this should be fixed now

fericsepi:
Going to squash and rebase to resolve the conflict.

PR #92, opened 2020-01-07
MAR-145 - improve get-user flow 

fericsepi:
- default email to 'email' field in sso details in case 'contact_email' is not present


dankingston:
All good for user emails going forward, but would be good to write a data migration so that existing users without an email (i.e. both of us), get their email populated. Something like:

`UserModel.objects.filter(email__isnull=True).update(email=F('username'))`

Not sure if email would be blank or null?

PR #98, opened 2020-03-12
MAR-101: Archive a barrier

dankingston:
* Add archived explanation field
* Add unarchived fields
* Return archived activity in the history API endpoint
* Allow filtering of barriers by the `archived` field


api/barriers/models.py
```diff
@@ -212,6 +212,14 @@ def submit_report(self, submitted_by=None):
         self.save()
         return self
 
+    @property
+    def archived_user(self):
```
fericsepi: May I suggest to rename this to `cleansed_archived_by`, when I read `archived_user` I think of a user object which was archived.

api/barriers/models.py
```diff
@@ -212,6 +212,14 @@ def submit_report(self, submitted_by=None):
...
+        return self._cleansed_username(self.archived_by)
+
+    @property
+    def unarchived_user(self):
```
fericsepi: Same goes for this.

api/core/models.py
```diff
@@ -78,3 +79,35 @@ def unarchive(self):
         self.archived_by = None
         self.archived_on = None
         self.save()
+
+
+class FullyArchivableModel(ArchivableModel):
```
fericsepi: Hmm why not update ArchivableModel instead of overrides?

Also why not call this `ArchivableMixin`, because you still need to use it together with BaseModel.
-- OR --
Call it ArchivableModel and inherit from BaseModel.

api/barriers/models.py
dankingston: Yeah, it's probably not the best name! I called it `archived_user` to be consistent with the existing properties (`created_user` and `modified_user`) which do the same thing. It would be good to keep the naming of everything consistent, so I should probably change those other fields as well - not sure if they're heavily referenced in the codebase or not!

api/core/models.py
dankingston: There's quite a few other models which already use `ArchivableModel` (`Assessment`, `TeamMember`, `Document`, `Interaction`), and I didn't want to change them (having choices of duplicate, not a barrier etc, isn't applicable to those models). I guess as those choices are specific to the Barrier, it should really be on that model.

And yeah, it's a bit weird inheriting from both `BaseModel` and `ArchivableModel` (this was existing code). `BaseModel` adds fields like `created_on`, `modified_on` etc. I think you're right, ArhcivableModel should really be a mixin.

I'll clean it up!

api/barriers/models.py
fericsepi: ah in that case, never mind :)

api/barriers/models.py
fericsepi: nvm

PR #101, opened 2020-03-20
MAR-228: Data workspace endpoint

dankingston:
* Fix broken serializer fields
* Add Hawk credentials for data workspace

### DIFF
--- a/api/dataset/serializers.py
+++ b/api/dataset/serializers.py
@@ -203,11 +203,11 @@ class BarrierDataSetSerializer(serializers.Serializer):
         Unpack companies json field into company names
         """
         if obj.companies:
-            return [company['name'] for company in obj.companies]
+            return [company['name'] for company in obj.companies if company.get('name')]
 
     def get_company_ids(self, obj):
         """
         Unpack companies json field into company ids
         """
         if obj.companies:
-            return [company['id'] for company in obj.companies]
+            return [company['id'] for company in obj.companies if company.get('id')]
--- a/api/dataset/serializers.py
+++ b/api/dataset/serializers.py
@@ -203,15 +203,11 @@ class BarrierDataSetSerializer(serializers.Serializer):
         Unpack companies json field into company names
         """
         if obj.companies:
-            companies_dict = json.loads(obj.companies)
-            return list(companies_dict.values())
-        return None
+            return [company['name'] for company in obj.companies]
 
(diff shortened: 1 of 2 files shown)


api/dataset/serializers.py
```diff
@@ -203,15 +203,11 @@ def get_company_names(self, obj):
         Unpack companies json field into company names
         """
         if obj.companies:
-            companies_dict = json.loads(obj.companies)
-            return list(companies_dict.values())
-        return None
+            return [company['name'] for company in obj.companies]
```
fericsepi: hmm, since it's JSON it's possible to have an entry without id, so it might be better to have a check for that
```python
return [company['name'] for company in obj.companies if company.get('name')]
```

api/dataset/serializers.py
```diff
@@ -203,15 +203,11 @@ def get_company_names(self, obj):
...
 
     def get_company_ids(self, obj):
         """
         Unpack companies json field into company ids
         """
         if obj.companies:
-            companies_dict = json.loads(obj.companies)
-            return list(companies_dict.keys())
-        return None
+            return [company['id'] for company in obj.companies]
```
fericsepi: same here
```python
return [company['id'] for company in obj.companies if company.get('id')]
```

config/settings/base.py
```diff
@@ -231,6 +231,10 @@
     DATAHUB_HAWK_KEY = os.environ.get('DH_HAWK_KEY')
     DATAHUB_HAWK_ALGORITHM = os.environ.get('DH_HAWK_ALGORITHM', 'sha256')
 
+    DATA_WORKSPACE_HAWK_ID = os.environ.get("DATA_WORKSPACE_HAWK_ID")
+    DATA_WORKSPACE_HAWK_KEY = os.environ.get("DATA_WORKSPACE_HAWK_KEY")
+    DATA_WORKSPACE_HAWK_ALGORITHM = os.environ.get("DATA_WORKSPACE_HAWK_ALGORITHM", "sha256")
```
fericsepi: will these need adding to the docker env file?

dankingston: They're not required for local development - it's just an extra set of hawk credentials, so it doesn't cause any errors etc not having them.
I've added the values on vault for dev/uat/prod so it should all work when we deploy.

PR #102, opened 2020-03-20
MAR-54 - unseen changes

fericsepi:
MAR-54:
- add last_seen_on field to barrier
- new model BarrierUserHits to track user-barrier view timestamps


api/barriers/models.py
```diff
@@ -220,6 +216,14 @@ def created_user(self):
     def modified_user(self):
         return self._cleansed_username(self.modified_by)
 
+    @property
+    def last_seen_by(self, user_id):
```
dankingston: Guessing it shouldn't be a property if it takes an argument? Is this used anywhere?

api/barriers/models.py
```diff
@@ -243,6 +247,16 @@ def save(
         )
 
 
+class BarrierUserHits(models.Model):
```
dankingston: Model names should really be singular - could we change it to `BarrierUserHit`?

dankingston:
Also nice to see some of the PEP8 stuff cleaned up!

api/barriers/models.py
fericsepi: ohyeah

fericsepi:
resolved conflicts and migrations

PR #104, opened 2020-03-26
MAR-144: Barrier history

dankingston:
* Barrier history tab
* Start tracking m2m fields (categories an documents) and populate them historically
* Rename `barrier_types` to `categories`


api/barriers/migrations/0045_auto_20200325_1458.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 2.2.10 on 2020-03-25 14:58
+
```
fericsepi: please rename the file and remove the comment - this wasn't generated by django

api/interactions/migrations/0006_auto_20200325_1637.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 2.2.10 on 2020-03-25 16:37
```
fericsepi: same here

PR #109, opened 2020-04-03
MAR-32 - tags

fericsepi:
MAR-32:
 - add model for tags and include them in metadata
 - removing eu_exit_related field as tags are favoured
 - utils to easily backup and restore local api DB (to help rapidly rollback DB should it be neccessary)
 - migrations to add brexit tag to all barriers where eu_exit_related field was 1
 - add ability to filter barriers by tags (list api)
 - add utility function to manage/adjust barrier tags
 - add optional tags field to barrier serializer


api/barriers/migrations/0054_convert_to_tags.py
```diff
@@ -0,0 +1,100 @@
...
+
+COVID_TAG_TITLE = "COVID-19"
+BREXIT_TAG_TITLE = "Brexit"
+
+DEFAULT_TAGS = (
+    {
+        "title": COVID_TAG_TITLE,
+        "description": (
+            "Barrier imposed because of the Coronavirus (COVID-19) pandemic. "
+            "For example restrictions related to medical supplies or food supply chains."
```
dankingston, line 12: This copy got updated to:

> Barrier imposed because of the COVID-19 (coronavirus) pandemic. For example restrictions related to medical supplies or food supply chains.

PR #110, opened 2020-04-06
MAR-216: Consistent status form

dankingston:
The status form during the add a barrier flow is now the full form (Open pending, Open in progress, Partially resolved, Fully resolved, Dormant) instead of the abbreviated form (Yes/Partially/No).

* Deprecate the fields `is_resolved`, `resolved_date` and `resolved_status` as these are no longer needed.
* Add a `draft` boolean field instead of using `status=0` to flag draft barriers, as barriers can now have a status whilst also being draft.
* Data migrations to populate `draft` and report stage statuses correctly
* Update barrier validation to handle `status` instead of `is_resolved` etc
* Update the tests to remove deprecated fields.


samuelhwilliams:
Hey 👋 Just a heads-up, this breaking change broke our pipeline using the barriers dataset endpoint this morning as we were still expecting a `resolved_date` field.

Not a complex update for us, so just FYI.

PR #112, opened 2020-04-08
MAR-216: Consistent status form

dankingston:
* Allow history classes to filter history by custom date
* Activity endpoint now returns items from before the `reported_on` date


api/barriers/history/base.py
```diff
@@ -67,15 +67,15 @@ def create(cls, field, new_record, old_record):
...
 
     @classmethod
-    def get_history(cls, barrier_id):
+    def get_history(cls, barrier_id, start_date=None):
         """ Fetch the history for a model """
         raise NotImplementedError
 
     @classmethod
-    def get_history_items(cls, barrier_id, fields=[]):
+    def get_history_items(cls, barrier_id, fields=[], start_date=None):
```
fericsepi, line 75: hmm, default for field should not be a mutable

api/barriers/views.py
```diff
@@ -529,28 +530,32 @@ class HistoryMixin:
...
-    def get_notes_history(self, fields=[]):
+    def get_notes_history(self, fields=[], start_date=None):
         return NoteHistoryFactory.get_history_items(
             barrier_id=self.kwargs.get("pk"),
             fields=fields,
+            start_date=start_date,
         )
 
-    def get_team_history(self, fields=[]):
+    def get_team_history(self, fields=[], start_date=None):
```
fericsepi, line 554: same in this file with `fields=[]`

PR #115, opened 2020-04-09
MAR-29: Reported terminology

dankingston:
Use 'reported' instead of 'created' or 'added' for barriers


api/collaboration/migrations/0004_change_creator_to_reporter.py
```diff
@@ -0,0 +1,31 @@
...
+def change_creator_to_reporter(apps, schema_editor):
+    TeamMember = apps.get_model('collaboration', 'TeamMember')
+    TeamMember.objects.filter(role="Barrier creator").update(role="Reporter")
+    HistoricalTeamMember = apps.get_model('collaboration', 'HistoricalTeamMember')
+    HistoricalTeamMember.objects.filter(role="Barrier creator").update(role="Reporter")
+
+
+def change_reporter_to_creator(apps, schema_editor):
+    TeamMember = apps.get_model('collaboration', 'TeamMember')
+    TeamMember.objects.filter(role="Reporter").update(role="Barrier creator")
```
fericsepi, line 15: shouldn't this be the opposite?
```python
TeamMember.objects.filter(role="Barrier creator").update(role="Reporter")
```

api/collaboration/migrations/0004_change_creator_to_reporter.py
```diff
@@ -0,0 +1,31 @@
...
+    HistoricalTeamMember = apps.get_model('collaboration', 'HistoricalTeamMember')
+    HistoricalTeamMember.objects.filter(role="Reporter").update(role="Barrier creator")
```
fericsepi, line 17: same here

PR #116, opened 2020-04-09
Add Brexit and COVID tag values to csv download

dankingston:


api/barriers/views.py
```diff
@@ -446,6 +446,8 @@ class BarriertListExportView(generics.ListAPIView):
             "reported_on": "Reported Date",
             "status_date": "Status Date",
             "modified_on": "Last updated",
+            "brexit": "Brexit",
```
fericsepi, line 449: why not use the constants here?

dankingston: Yeah, I guess that would make sense, although these column titles might not always necessarily be the same as the tag e.g. "Is Brexit related?", "Is Covid-19 related?"

PR #123, opened 2020-04-15
Mar 270/trade direction

fericsepi:


api/barriers/models.py
```diff
@@ -48,7 +49,7 @@ class BarrierManager(models.Manager):
     """ Manage barriers within the model, with draft=False """
 
     def get_queryset(self):
-        return super(BarrierManager, self).get_queryset().filter(draft=False)
+        return super().get_queryset().filter(Q(draft=False) & Q(archived=False))
```
dankingston, line 52: I think this might mess up the barrier search. We need to be able to search for archived barriers (i.e. when you select "show only archived"), so this could mean archived barriers aren't returned by the API any more.

fericsepi: ah right - nicely spotted
I think I had a test failing because of this - let me update the PR

PR #128, opened 2020-04-23
MAR-63 - team member management

fericsepi:


api/barriers/history/team_members.py
```diff
@@ -21,12 +21,10 @@ def create_history_items(cls, new_record, old_record, fields=()):
         if new_record.history_type == "+":
             return [TeamMemberHistoryItem(new_record, None)]
         if new_record.history_type == "~":
-            if new_record.user == old_record.user:
```
dankingston, line 24: I think we need to keep this line, but change it to:
`if new_record.id == old_record.id:`

Otherwise it could compare changes from one team member record against a different team member record.

api/barriers/views.py
```diff
@@ -419,6 +409,19 @@ def team_barriers(self, queryset, name, value):
             ).distinct()
         return queryset
 
+    def member_filter(self, queryset, name, value):
+        if value:
+            member = get_object_or_404(TeamMember, pk=value)
+            return queryset.annotate(
```
dankingston, line 415: This can probably be simplified a bit:
`return queryset.filter(barrier_team__pk=value)`

api/collaboration/migrations/0005_team_members_overhaul.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def convert_to_contributors(apps, schema_editor):
+    """
+    Convert all existing team members to contributors
+    apart from reporters and potential owners.
+    """
+    TeamMember = apps.get_model('collaboration', 'TeamMember')
+    HistoricalTeamMember = apps.get_model('collaboration', 'HistoricalTeamMember')
+    TeamMember.objects.exclude(role__in=(REPORTER, OWNER)).update(role=CONTRIBUTOR)
```
dankingston, line 68: Could this lead to duplicate contributors? If they've made changes to the barrier they will already be listed as a contributor (from line 27 above).

fericsepi: `convert_to_contributors` is not adding extra members, it's just converting existing (non Owner and non Reporter) members to  Contributors.

api/barriers/views.py
fericsepi: I think I have tried that first but unfortunately, that is not the same :/

api/barriers/history/team_members.py
fericsepi: History items are not created then that if this is in place (when changing the owner) :<

api/barriers/views.py
dankingston: Sorry - got it slightly wrong, but `return queryset.filter(barrier_team__user=member.user)` should do the same (just tried and it works for me). But doesn't really matter if they're both doing the same thing.

api/collaboration/migrations/0005_team_members_overhaul.py
dankingston: As an example...
`User 1` has made edits to a barrier, and listed as a having `Some role`.
The first part of the migration would add `User 1` as a `Contributor` (because they have made edits).
The second part of the migration would convert `Some role` to `Contributor`, leading to them being listed as a contributor twice.

I found an example where it looks like this has happened:
https://market-access-dev.london.cloudapps.digital/barriers/04757c76-8b81-41b0-b2c7-6c67eb78ef05/team/

api/barriers/history/team_members.py
dankingston: I think it was just history items were not being created when this was comparing `record.user` rather than `record.id`.

Without checking the ids match it can end up comparing two totally different `TeamMember` objects and try to diff them, which would lead to weird entries in the history tab. We only want to compare like with like (a TeamMember object with a previous version of itself i.e. with the same id).

It does work ok at the moment though because for each barrier, only ever one TeamMember can be changed (the owner), so you won't get mixed up history items. But in the past people were able to change different roles, so those edits will show up weird in the history tab.

fericsepi: it's the same behaviour when we compare `record.id`-s as well
the record is not being created and therefore the event doesn't show up in the history tab

api/collaboration/migrations/0005_team_members_overhaul.py
fericsepi: Where do you read that? :)
The flow is:
If `User 1` already has a corresponding `Member` record (with any role) the first part will just skip it.
Then the second part will ensure that its role is Contributor (but only if the role was not a Reporter or Owner).

api/barriers/views.py
fericsepi: ah yeah 🤦

fericsepi:
rebased to solve conflict

PR #130, opened 2020-04-28
MAR-269: WTO

dankingston:
* `WTOProfile` model - in a separate model to avoid bloating the `Barrier` model and for better data normalisation
* WTO history items
* WTO search filters
* WTO committee models and data migration (ids should be the same as in the mock metadata on the frontend to avoid problems)


api/wto/models.py
```diff
@@ -0,0 +1,94 @@
...
+
+from api.interactions.models import Document
+
+from simple_history.models import HistoricalRecords
+
+MAX_LENGTH = settings.CHAR_FIELD_MAX_LENGTH
+
+
+class WTOCommitteeGroup(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid4)
```
fericsepi, line 15: don't mind to leave `default` explicitly declared there but the default for UUIDField is uuid4 :)

fericsepi: Same goes for the other models in this file

PR #154, opened 2020-06-05
MAR-247: Add correlation ids to logging

dankingston:
Use `django_log_formatter_ecs` to add correlation ids to logs

### DIFF
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -3,6 +3,7 @@ from .base import *
 
 from django_log_formatter_ecs import ECSFormatter
 
+# SECURITY WARNING: don't run with debug turned on in production!
 DJANGO_ENV = 'dev'
 SSO_ENABLED = True
 
--- a/config/settings/dev.py
+++ b/config/settings/dev.py
@@ -1,4 +1,30 @@
+import sys
 from .base import *
+
+from django_log_formatter_ecs import ECSFormatter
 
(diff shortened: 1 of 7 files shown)


config/settings/production.py
```diff
@@ -1,6 +1,31 @@
...
+            "stream": sys.stdout,
+        },
+    },
+    "loggers": {
+        "": {
+            "handlers": ["ecs"],
+            "level": DJANGO_LOG_LEVEL
+        }
+    },
+}
```
fericsepi, line 31: Looks like you're overriding the base LOGGING config here (from `base.py`) and we would loose the json formatter?

config/settings/production.py
```diff
@@ -1,6 +1,31 @@
+import sys
 from .base import *
 
-# SECURITY WARNING: don't run with debug turned on in production!
+from django_log_formatter_ecs import ECSFormatter
+
```
fericsepi, line 5:
```suggestion

# SECURITY WARNING: don't run with debug turned on in production!
```

config/settings/dev.py
```diff
@@ -1,4 +1,30 @@
+import sys
 from .base import *
 
+from django_log_formatter_ecs import ECSFormatter
+
```
fericsepi, line 5:
```suggestion

# SECURITY WARNING: don't run with debug turned on in production!
```

PR #171, opened 2020-07-07
MAR-454 - add a new field to Assessment so explanation for commercial…

fericsepi:
… value can be saved


api/assessment/models.py
```diff
@@ -74,6 +74,7 @@ class Assessment(ArchivableMixin, BarrierRelatedMixin, BaseModel):
     value_to_economy = models.BigIntegerField(null=True)
     import_market_size = models.BigIntegerField(null=True)
     commercial_value = models.BigIntegerField(null=True)
+    commercial_value_explanation = models.TextField(null=True, default=None)
```
dankingston, line 77: The Django docs recommend using `null=False, blank=True` for `TextField`:

> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. 

Although that's not really been stuck to in the rest of the project.

PR #176, opened 2020-07-28
Mar 213/publish barriers user permissions

fericsepi:


dankingston:
We should really check for the underlying permissions rather than the Group the user belongs to. Potentially the permissions within a group could change and there could be other groups which are allowed to do things.

Checking for the permissions rather than the group should actually simplify things i.e. something like:
```
class CanPublish(BasePermission):
    def has_permission(self, request, view):
        return request.user.has_perm("barriers.publish_barrier")
```

 The underlying permissions are:

```
Sifter:
    change_barrier_public_eligibility

Editor:
    change_barrier_public_eligibility
    change_publicbarrier
    mark_barrier_as_ready_for_publishing

Publisher:
    change_barrier_public_eligibility
    change_publicbarrier
    mark_barrier_as_ready_for_publishing
    publish_barrier
```

PR #177, opened 2020-07-28
MAR-503 - add status_date to the serializer so it appears on the CSV

fericsepi:


api/barriers/serializers.py
```diff
@@ -244,6 +246,9 @@ def get_status(self, obj):
             status = f"{status} ({sub_status_dict.get(obj.sub_status, 'Unknown')})"
         return status
 
+    def get_status_date(self, obj):
+        return obj.status_date
```
dankingston, line 250: Will the date be formatted ok? Other fields seem to use this for formatting dates:
`modified_on = serializers.DateTimeField(format="%Y-%m-%d")`

fericsepi: it's throwing an error... rightly so because status_date is a DateField
fixup inc

PR #186, opened 2020-08-12
MAR-519: Cache history items

dankingston:
* `CachedHistoryItem` model - allows querying on what field was changed and when (instead of calculating by comparing historical instances)
* Management command to build the history cache
* Use the `post_create_historical_record` signal to add a `CachedHistoryItem` as the change occurs
* Move history classes into separate app
* Move history querying into a `Manager` class instead of in the view
* Move `BarrierInstance.wto_profile` `OneToOneField` to `WTOProfile.barrier` - now more consistent with other models (`Assessment`, `PublicBarrier`) and means the wto profile object has a barrier when the signal is called.
* Add `status_history` to data workspace endpoint by querying `CachedHistoryItem` for status changes
* Retire old data workspace endpoint


tests/barriers/test_history.py
```diff
@@ -871,3 +872,123 @@ def test_history_endpoint(self):
...
+class TestCachedHistoryItems(APITestMixin, TestCase):
+    fixtures = ["categories", "documents", "users", "barriers"]
+
+    @freeze_time("2020-03-02")
+    def setUp(self):
+        self.barrier = BarrierInstance.objects.get(
+            pk="c33dad08-b09c-4e19-ae1a-be47796a8882"
+        )
+        self.barrier.save()
+        self.assessment = Assessment.objects.create(
```
fericsepi, line 886: why not use `AssessmentFactory`?

tests/barriers/test_history.py
```diff
@@ -871,3 +872,123 @@ def test_history_endpoint(self):
...
+            barrier=self.barrier,
+            impact="LOW",
+            explanation="Some explanation",
+            value_to_economy=10000,
+            import_market_size=20000,
+            commercial_value=30000,
+            export_value=40000,
+        )
+        self.note = Interaction.objects.create(
```
fericsepi, line 895: I would create a factory for Interactions

PR #196, opened 2020-08-24
MAR-362: European Union

dankingston:
* Barriers can now have a `trading_bloc` instead of a country
* `caused_by_trading_bloc` field to signal whether a barrier is caused by its country's trading bloc
* Add trading blocs to search filters
* Add trading blocs to metadata
* Update `PublicBarrier` model to reflect that location can be a country or trading bloc


api/barriers/serializers/csv.py
```diff
@@ -175,10 +176,15 @@ def get_sectors(self, obj):
         else:
             return "N/A"
 
-    def get_country(self, obj):
-        country = get_country(str(obj.export_country))
-        if country:
-            return country.get("name")
+    def get_location(self, obj):
```
fericsepi, line 179: I would suggest to either move this logic to the models or to helpers - like get_barrier_location that would return something like
```python
{
  "id": "some id"
  "name": "Some name",
  "type": "country or t-bloc", 
}
```

Just in case we need deal with the same situation elsewhere in the code.

fericsepi (approved):
the PR looks good otherwise - I don't want to block it if you have no time to move it now

PR #204, opened 2020-09-22
MAR-535: Public resolution status

dankingston:
* Add `status_date` to `PublicBarrier` model
* Add `is_resolved` and `status_date` fields to public barrier serializer


api/barriers/models.py
```diff
@@ -736,11 +762,15 @@ def internal_trading_bloc_changed(self):
 
     @property
     def internal_location(self):
-        return self.barrier.location
+        return get_location_text(
```
fericsepi, line 765: `self.barrier.location` is already using the helper so there's no need to call `get_location_text` here

api/barriers/models.py
```diff
@@ -736,11 +762,15 @@ def internal_trading_bloc_changed(self):
...
+            country_id=self.barrier.export_country,
+            trading_bloc=self.barrier.trading_bloc,
+            caused_by_trading_bloc=self.barrier.caused_by_trading_bloc,
+        )
 
     @property
     def internal_location_changed(self):
-        return self.barrier.location != self.location
+        return self.internal_location != self.location
```
fericsepi, line 773: it does the same, but reads better 👍

api/barriers/models.py
fericsepi: as discussed - not surfacing admin areas, and self.barrier.location is using that

fericsepi (approved):

PR #220, opened 2020-10-14
MAR-611 - add ability to walk s3 files of published data, versioned f…

fericsepi:
…ile class to help determine the latest file on s3 and to be able to determine what the next version should be


api/barriers/public_data.py
```diff
@@ -6,18 +6,83 @@
...
+            revision = self.next_revision(0)
+        return f"v{settings.PUBLIC_DATA_MAJOR}.{settings.PUBLIC_DATA_MINOR}.{revision}"
+
+    @property
+    def version_as_float(self):
+        """
+        To be able to compare file versions easily
+        :return: FLOAT - i.e.: 10.12345
+        """
+        return float(f"{self.major}{self.minor}.{self.revision}")
```
dankingston, line 64: I guess minor shouldn't go above 9, otherwise major=2, minor=1 would be a lower version than major=1, minor=10 ?

fericsepi: good point

dankingston (approved):

PR #226, opened 2020-11-03
MAR-516: Tidy up modelling

dankingston:
Tidy up model and field names:
* Rename `BarrierInstance` to `Barrier`
* Rename `barrier_title` to `title`
* Rename `export_country` to `country`
* Rename `country_admin_areas` to `admin_areas`
* Rename `problem_status` to `term`


dankingston:
Agree with most of the Foxy Doctor's comments, but there's a separate ticket for correcting null/blank values (MAR-672)
This ticket is purely about renaming fields/models.

fericsepi (approved):

PR #249, opened 2021-02-03
bump python version to version installed in Dockerfile

jamespacileo:
When running market-access-api locally the web container failed to load as the installed python version in the container seems to be a version ahead compared to the local entrypoint.sh

I've updated the binary pointed to from specific python3.6 to generic python, which always point to the correct version


rossmiller (approved):
LGTM!

PR #252, opened 2021-02-10
Mar 694 - Public barrier list API enhancements

jamespacileo:


api/barriers/serializers/public_barriers.py
```diff
@@ -183,20 +188,30 @@ def get_internal_code(self, obj):
     def get_internal_id(self, obj):
         return obj.barrier_id
 
+    def get_internal_government_organisations(self, obj):
+        return OrganisationSerializer(obj.barrier.organisations, many=True).data
+
     def get_latest_note(self, obj):
         try:
-            note = obj.notes.latest("created_on")
+            # We need to perform Python sorting instead of SQL
```
rossmiller, line 196: Could you use .first() here?

rossmiller (commented):
One comment to address.

rossmiller (approved):
LGTM!

api/barriers/serializers/public_barriers.py
jamespacileo: Had to avoid `first()` unfortunately as it overrides prefetct_related and triggers a new query

PR #254, opened 2021-02-16
Mar 694

jamespacileo:


api/barriers/models.py
```diff
@@ -573,12 +617,13 @@ class PublicBarrier(FullyArchivableMixin, BaseModel):
     This table should not be exposed to the public however only to the DMAS frontend which requires login.
     Transfer the data to a flat file or another service which can safely expose the data.
     """
+
     id = HashidAutoField(
-        primary_key=True,
-        min_length=6,
-        alphabet="0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        primary_key=True, min_length=6, alphabet="0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
cad106uk, line 622: Why is the alphabet manually coded here? If this alphabetic is needed is should be imported from the inbuilt `string` library

cad106uk (approved):
I just have some random question. I cann't see anything obviously wrong here.

cad106uk (commented):
looks good

PR #255, opened 2021-02-18
Mar 837

jamespacileo:


tests/barriers/test_public_barriers.py
```diff
@@ -161,7 +166,7 @@ def test_pb_list_organisation_filter_with_multiple_values(self):
 
     def test_pb_list_status_filter(self):
 
-        barriers = {}
+        barriers: Dict[int, Barrier] = {}
```
cad106uk, line 169: Should this be?

```python
barriers: dict[PublicBarrierStatus, Barrier] = {}
```

tests/barriers/test_public_barriers.py
```diff
@@ -204,6 +209,56 @@ def get_list_for_status(status_code):
...
+                        "barrier__cached_history_items__date"
+                    ),
+                    then=Value(True),
+                ),
+                default=Value(False),
+                output_field=BooleanField(),
+            ),
+        )
+
+        summary_change = [
```
cad106uk, line 255: Personally, I like writing 

```python
summary_change = next((query_change for query_change in pb_query if query_change.change == "barrier.summary"))
```

The reason is, with the list comprehension the entire list has to be generated and put into memory then only the first element is taken. With the example, an iterator is made and only the first element is generated and returned.

This is only really an advantage if we are dealing with very long lists. However, it does make the execution profile of the process much more predictable when scaling to much larger data sets. Basically, I am just being a smart arse here unless you think there is a good chance the list you are dealing with could jump or grow to a much bigger size

cad106uk (approved):

PR #259, opened 2021-02-23
MAR-720

jamespacileo:


api/assessment/automate/calculator.py
```diff
@@ -163,43 +163,47 @@ def calculate(
         # 1. Partner country imports of affected products from UK
         partner_from_uk = avgtrade(
             df=affected_products_df,
-            reporter_input=country1,
-            partner_input=country2,
+            reporter_input=get_comtrade_country_name(country1),
```
cad106uk, line 166: Why are you adding `get_comtrade_country_name` everywhere `avgtrade` is called? Just update `avgtrade` function,  fewer code changes and everything is called consistently. (This will work with the "World" option as well)

api/assessment/serializers.py
```diff
@@ -117,18 +115,13 @@ class Meta:
...
-                    commodity_codes=commodity_codes,
-                    product=barrier.product,
-                    country1=barrier.country_name,
+                data = calculate_barrier_economic_assessment(
+                    validated_data["barrier_id"]
                 )
+                validated_data["automated_analysis_data"] = data
             except ComtradeError as e:
                 raise ValidationError(e)
-            validated_data["automated_analysis_data"] = data
```
cad106uk, line 131: Why has this dictionary assignment been moved inside the try block? It will only hide errors.

cad106uk (changes requested):

cad106uk (changes requested):
There are no unittests for `get_comtrade_country_name` and `calculate_barrier_economic_assessment`

api/assessment/serializers.py
jamespacileo: data is only assigned to inside the block, if data fails to be instantiated an exception is already being raised, thought it might make sense to keep the assignment to validated_data is the same block

PR #262, opened 2021-03-03
MAR-811

jamespacileo:


api/assessment/automate/calculator.py
```diff
@@ -30,19 +30,18 @@ def get_year_range(self, country1, country2, year=None):
...
-        end_year = self.client.get_valid_year(
+        valid_years = self.client.get_valid_years(
             year,
             get_comtrade_country_name(country1),
             get_comtrade_country_name(country2),
         )
-        start_year = end_year - 2
 
-        if not use_most_recent and end_year != year:
+        if not use_most_recent and valid_years[0] != year:
```
cad106uk, line 39: Do you need to check if `valid_years` empty?

api/assessment/automate/calculator.py
```diff
@@ -390,11 +384,11 @@ def calculate(
 
         # Market share of UK exports of products affected (UK to partner / UK to world)
         # Do we still need this? Not in current MAB assessment template.
-        share_exports_affected = percent_range(
-            partner_from_uk / world_from_uk,
-            uk_to_partner / uk_to_world,
-            decimal_places=1,
-        )
+        # share_exports_affected = percent_range(
```
cad106uk, line 387: Don't add commented out code. Just delete and make a commit recording that you have deleted the code here.

If it is significant code you can add a git tag pointing to where the code was deleted.

api/assessment/automate/calculator.py
```diff
@@ -430,8 +424,8 @@ def calculate(
             "version": self.version,
             "commodity_codes": commodity_codes,
             "product": product,
-            "start_year": start_year,
-            "end_year": end_year,
+            "start_year": years[-1],
```
cad106uk, line 427: 1. Does it matter if `start_year` and `end_year` are the same data?
2. Why is the start date and the end of the `years` list? Isn't that backwards? Why (should) the list be sorted this way?
3. Why does this still have start date and end date where everywhere else it has been replaced with `years`?

api/assessment/automate/comtrade.py
```diff
@@ -206,4 +203,9 @@ def get_valid_year(self, target_year, country1, country2):
             years = [item["yr"] for item in data]
 
             if len(years) == 4 and all(y == year for y in years):
-                return year
+                valid_years.append(year)
+
+            if len(valid_years) == 3:
```
cad106uk, line 208: - This should be on line 204 to stop the loop as soon as we know we have a 'bad' value,  or at least a condition where we should stop the loop
- Could the condition here be `len(valid_years) != 4)`

api/assessment/automate/comtrade.py
```diff
@@ -206,4 +203,9 @@ def get_valid_year(self, target_year, country1, country2):
...
+                break
+
+        return valid_years
```
cad106uk, line 211: Should this method be a generator? Instead of returning a list, use `yield year` on line 206 instead of building a list?

cad106uk (changes requested):
At the very least the commented out code has to go away.

api/assessment/automate/comtrade.py
jamespacileo: Upgrading the method to a generator is a cool idea, but there are the results would have to be resolved into a list in the only location it is used. I'm happy to upgrade if we think it's a must

api/assessment/automate/comtrade.py
jamespacileo: My plan was to return the years as soon as we have 3 of them, or let the loop end and return the list we have.

I believe that line 208 is a valid location for the check, and will return the values without having to go into the next loop.

Also, apologies, but I'm not sure why we'd check for `len(valid_years) != 4)`

Thanks for your feedback

api/assessment/automate/calculator.py
jamespacileo: I think this is a good idea, will pop it in, thanks

api/assessment/automate/calculator.py
jamespacileo: yeah no problem, let's delete this completely

api/assessment/automate/calculator.py
jamespacileo: The original code works backwards, but we can sort the list above so that it creates less confusion.

Frontend needs to know the initial year and ending year in which the data was collected, but we can supply the list of years collected for future use.

cad106uk: Coolio, a comment here saying the `start_date` is at the end of the list is enough here.

jamespacileo: awesome thanks!

api/assessment/automate/comtrade.py
cad106uk: There is no need to change the method. My default thought is to make lazy executables and only take resources as late as possible for a short a time as possible. Honestly, though it is rare that it is necessary to do this.

cad106uk (approved):

PR #266, opened 2021-03-24
MAR-822 - Final set of Mention features

jamespacileo:


api/interactions/views.py
```diff
@@ -1,19 +1,74 @@
...
+from django.db import transaction
+from django.shortcuts import get_object_or_404
+from rest_framework import generics, viewsets
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+
+
+class ExcludeFromNotificationsView(viewsets.ViewSet):
+    def retrieve(self, request, pk=None):
+        user_has_exclusion = (
```
cad106uk, line 31: instead, use  `ExcludeFromNotifications.objects.filter(excluded_user=request.user).exists()`

and get the boolean directly

api/interactions/views.py
```diff
@@ -1,19 +1,74 @@
...
+            )
+            data.is_valid()
+            return Response(data.validated_data)
+
+    def create(self, request):
+        # if request.user.is_anonymous():
+        #     raise Exception("User is anonymous")
+        ExcludeFromNotifications.objects.get_or_create(
+            excluded_user=request.user,
+            defaults={
```
cad106uk, line 54: Why are you using defaults here?

cad106uk (changes requested):
All these migration files should be merged using the Makefile command django-makemigrations


My patch will be coming after lunch

PR #279, opened 2021-04-21
Mar 859

jamespacileo:


api/barriers/filters.py
```diff
@@ -202,16 +205,27 @@ def text_search(self, queryset, name, value):
...
+            return (
+                queryset.filter(
+                    Q(barrier_team__user=current_user) & Q(barrier_team__archived=False)
+                )
+                .exclude(created_by=current_user)
+                .distinct()
+            )
+        return queryset
+
+    def my_draft_barriers(self, queryset, name, value):
```
cad106uk, line 224: What is the value value? Is this a flag being sent to this function? Why?

Is this function being used anywhere?

cad106uk (changes requested):

api/barriers/filters.py
jamespacileo: good catch, removed it.

cad106uk (approved):

PR #286, opened 2021-05-07
Mar 952 an authnentication broker

cad106uk:


api/user/backend.py
```diff
@@ -0,0 +1,65 @@
...
+            profile_obj.sso_email_user_id = raw_profile["email_user_id"]
+            profile_obj.user.email = raw_profile.get("contact_email") or email
+            # contact emails ?
+            profile_obj.user.first_name = raw_profile[
+                "first_name"
+            ]  # might change over time
+            profile_obj.user.last_name = raw_profile[
+                "last_name"
+            ]  # might change over time
+            profile_obj.user.is_active = True
```
rossmiller, line 38: Why are you setting is_staff and is_active here?

api/user/backend.py
```diff
@@ -0,0 +1,65 @@
...
+            profile = get_profile(client)
+            return self.verify_user_object(profile)
+        return None
+
+    @staticmethod
+    def verify_user_object(raw_profile) -> User:
+        email = raw_profile["email"]
+
+        def _build_profile(profile_obj):
+            profile_obj.sso_user_id = raw_profile["user_id"]
```
rossmiller, line 28: Could we call it "profile" rather than "profile_obj"?

api/user/backend.py
```diff
@@ -0,0 +1,65 @@
...
+        if has_valid_token(client):
+            profile = get_profile(client)
+            return self.verify_user_object(profile)
+        return None
+
+    @staticmethod
+    def verify_user_object(raw_profile) -> User:
+        email = raw_profile["email"]
+
+        def _build_profile(profile_obj):
```
rossmiller, line 27: This would be better as a function of the class

rossmiller (changes requested):
Looks good. I've made some suggestions.

api/user/backend.py
cad106uk: Hmm, well, this is never meant to be used anywhere else or by anyone else, which is why I use an inline function.

Personal whenever I see a static method I always think "this should be a function". The only reason I can think of to use a static method is when you cannot know where the class in going to be imported. Personally, I would be happier if both these methods where functions.

I have made the change you asked for, I just grumbled a bit about it.

jamespacileo (approved):

rossmiller (approved):
LGTM!

PR #291, opened 2021-05-11
Mar 908

jamespacileo:


api/barriers/models.py
```diff
@@ -1188,9 +1256,7 @@ def location_filter(self, queryset, name, value):
             if "country_trading_bloc" in self.data:
                 trading_bloc_countries = []
                 for trading_bloc in self.data["country_trading_bloc"].split(","):
-                    trading_bloc_countries += (
-                        metadata_utils.get_trading_bloc_country_ids(trading_bloc)
-                    )
+                    trading_bloc_countries += metadata_utils.get_trading_bloc_country_ids(trading_bloc)
```
cad106uk, line 1259: A more Pythonic way of writing this.

```python

trading_bloc_countries = [
    metadata_utils.get_trading_bloc_country_ids(trading_bloc)
    for trading_bloc in self.data["country_trading_bloc"].split(",")
]

```

api/barriers/signals/handlers.py
```diff
@@ -67,6 +78,51 @@ def public_barrier_categories_changed(sender, instance, action, **kwargs):
             instance.save()
 
 
+def public_barrier_light_touch_reviews_changed(
+    sender, instance: PublicBarrierLightTouchReviews, **kwargs
+):
+    if hasattr(instance, "light_touch_reviews_history_saved"):
+        historical_instance = HistoricalPublicBarrier.objects.filter(
```
cad106uk, line 85: We are getting the latest row here shouldn't there be an order by in this select to make sure we are getting the correct row?

api/barriers/signals/handlers.py
```diff
@@ -67,6 +78,51 @@ def public_barrier_categories_changed(sender, instance, action, **kwargs):
...
+
+
+def public_barrier_content_update(
+    sender, instance: PublicBarrier, update_fields, **kwargs
+):
+    """
+    When public barrier summary or title is changed remove content team approval and
+    flag that content has changed since last approval
+    """
+    try:
```
cad106uk, line 102: I find this to be more readable

```python
 
    pb_filter = PublicBarrier.objects.filter(id=instance.id)
    if not pb_filter.exist():
        return
    previous_instance = pb_filter.first()

    has_public_content_changed = (instance.title != previous_instance.title) or (
        instance.summary != previous_instance.summary
    )

    if not has_public_content_changed:
        return 

    try:
        light_touch_reviews: PublicBarrierLightTouchReviews = (
            instance.light_touch_reviews
        )
    except PublicBarrier.light_touch_reviews.RelatedObjectDoesNotExist:
        light_touch_reviews = PublicBarrierLightTouchReviews.objects.create(
            public_barrier=instance
        )
    if not light_touch_reviews.content_team_approval:
        return

    light_touch_reviews.content_team_approval = False
    light_touch_reviews.has_content_changed_since_approval = True
    light_touch_reviews.save()
```

api/barriers/views.py
```diff
@@ -834,3 +840,69 @@ def unpublish(self, request, *args, **kwargs):
...
         return r
+
+    @action(methods=["post"], detail=True, permission_classes=())
+    def mark_approvals(self, request, *args, **kwargs):
+        public_barrier = self.get_object()
+        light_touch_reviews: PublicBarrierLightTouchReviews = (
+            public_barrier.light_touch_reviews
+        )
+        serializer = LightTouchApprovalSerializer(data=request.data.get("approvals"))
+        if serializer.is_valid():
```
cad106uk, line 851: This is a guard so should be written as 

```python
if not serializer.is_valid():
    return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```

The rest of the code need not be indented.

api/barriers/views.py
```diff
@@ -834,3 +840,69 @@ def unpublish(self, request, *args, **kwargs):
...
+    def enable_hm_trade_commissioner_approvals(self, request, *Args, **kwargs):
+        public_barrier = self.get_object()
+        light_touch_reviews: PublicBarrierLightTouchReviews = (
+            public_barrier.light_touch_reviews
+        )
+
+        serializer = LightTouchReviewsEnableHMTradeCommissionerSerializer(
+            data=request.data
+        )
+        if serializer.is_valid():
```
cad106uk, line 885: Again this is a guard

cad106uk (changes requested):
All good, just a couple of changes.

api/barriers/models.py
jamespacileo: thanks a ton! Yeah this is much better!

api/barriers/signals/handlers.py
jamespacileo: Hi I believe the simple-history is already ordered by the history_date, I've added the name of the field in latest. Which you are right is good practice.

api/barriers/signals/handlers.py
jamespacileo: Thank you for finding the time to refactor this, agree I should be using the guards as such. Looks more readable.

api/barriers/views.py
jamespacileo: fixed!

api/barriers/views.py
jamespacileo: fixed!

PR #313, opened 2021-08-18
Mar 1051

jamespacileo:


api/barriers/models.py
```diff
@@ -1201,6 +1162,14 @@ def priority_filter(self, queryset, name, value):
         else:
             return queryset.filter(priority__in=priorities)
 
+    def has_action_plan_filter(self, queryset, name, value):
+        if value:
```
cad106uk, line 1166: This is a guard and should be written as 

```python
if not value:
    return queryset
```

api/barriers/models.py
```diff
@@ -1201,6 +1162,14 @@ def priority_filter(self, queryset, name, value):
...
+            from api.action_plans.models import ActionPlanMilestone
```
cad106uk, line 1167: Why is this imported here and not at the top of the file

api/history/manager.py
```diff
@@ -63,15 +60,16 @@ def get_public_activity(cls, public_barrier, use_cache=False):
 
     @classmethod
     def get_full_history(cls, barrier, ignore_creation_items=False, use_cache=False):
-        if use_cache:
-            return cls.get_full_history_cached(barrier, ignore_creation_items)
+        # if use_cache:
```
cad106uk, line 63: This committed out code should just be deleted

config/settings/base.py
```diff
@@ -252,6 +252,7 @@
     "PAGE_SIZE": 400,
     "DEFAULT_AUTHENTICATION_CLASSES": [
         "oauth2_provider.contrib.rest_framework.OAuth2Authentication",
+        "rest_framework.authentication.SessionAuthentication",
```
cad106uk, line 255: Why is a new authentication system being added here?

cad106uk (changes requested):
some feedback

api/barriers/models.py
jamespacileo: circular dependency

api/barriers/models.py
jamespacileo: done

api/history/manager.py
jamespacileo: done

config/settings/base.py
jamespacileo: removed

cad106uk (approved):

PR #314, opened 2021-08-19
Mar 1051 - patch

jamespacileo:


api/barriers/models.py
```diff
@@ -1163,12 +1163,13 @@ def priority_filter(self, queryset, name, value):
...
     def has_action_plan_filter(self, queryset, name, value):
-        if value:
-            from api.action_plans.models import ActionPlanMilestone
+        if not value:
+            return queryset
 
-            milestones = ActionPlanMilestone.objects.all()
-            return queryset.filter(action_plans__milestones__in=milestones)
-        return queryset
+        from api.action_plans.models import ActionPlanMilestone
```
kevincarrogan, line 1169: Why is this import happening here?

api/barriers/models.py
```diff
@@ -1163,12 +1163,13 @@ def priority_filter(self, queryset, name, value):
...
+
+        milestones = ActionPlanMilestone.objects.all()
+        return queryset.filter(action_plans__milestones__in=milestones)
```
kevincarrogan, line 1172: This kind of confuses me. What's the actual intention here?

jamespacileo: it's just swapping the flow so the main logic isn't nested in an if

api/barriers/models.py
jamespacileo: import causes a circular dependency, possibly we could move this whole class outside into a third file and that would resolve this

we can either refactor this now or add a TODO comment and a backlog ticket to solve this when we have more time

kevincarrogan (approved):

api/barriers/models.py
jamespacileo: yes it's returning the list of items that have at least 1 milestone on the action plan model

PR #321, opened 2021-09-14
MAR-1097 - Patch security issues + pre commit checks

jamespacileo:


docker/local/Dockerfile
```diff
@@ -63,15 +63,19 @@ ENV LC_ALL en_GB.utf-8
...
+    POETRY_NO_INTERACTION=1 \
+    POETRY_VIRTUALENVS_CREATE=false
 
 WORKDIR /usr/src/app
 
 # Poetry commands reference https://python-poetry.org/docs/cli/#run
 COPY pyproject.toml poetry.lock ./
 RUN pip install --upgrade pip \
- && pip install poetry==$POETRY_VERSION \
- && poetry lock \
```
jamespacileo, line 74: removed creation of new lock as we want to preserve this between installs

kevincarrogan (approved):

PR #327, opened 2021-10-29
MAR-1049 - Importing commodity codes with old classification codes + fix import script arguments and duplication

jamespacileo:


docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
       - ./docker/local/comtrade_test_db:/docker-entrypoint-initdb.d:delegated
 
   db:
-    image: postgres
+    image: postgres:13
```
jamespacileo, line 16: version 14+ not compatible with current database

api/commodities/management/commands/import_commodities.py
```diff
@@ -23,17 +23,21 @@ def import_commodities(self, csv_file, version):
...
+                        code=line["Commodity code"].ljust(10, "0"),
+                        suffix=line["Suffix"],
+                        level=line["HS Level"],
+                        indent=line["Indent"],
+                        description=line["Description"],
+                        is_leaf=(line["Suffix"] == "80"),
+                        sid=line["SID"],
+                        parent_sid=line["Parent SID"] or None,
+                        classification=line.get("Classification", "H5"),
+                    )
```
kevincarrogan, line 40: You could use `exists` instead of `count` here but even better you could probably just use `get_or_create` with `defaults` and achieve the same thing.

kevincarrogan (approved):

PR #335, opened 2021-11-26
Mar 1118

jamespacileo:


api/barriers/models.py
```diff
@@ -1625,3 +1632,91 @@ def status_filter(self, queryset, name, value):
...
+
+        recipient_emails = settings.SEARCH_DOWNLOAD_APPROVAL_REQUEST_EMAILS
+
+        client = NotificationsAPIClient(settings.NOTIFY_API_KEY)
+        group_id = Group.objects.get(
+            name=settings.APPROVED_FOR_BARRIER_DOWNLOADS_GROUP_NAME
+        ).id
+        user_group_approval_path = f"/users/add/?group={group_id}".format(
+            group_id=group_id
+        )
```
kevincarrogan, line 1698: Is this supposed to be an f-string and using format at the same time?

api/user/migrations/0024_create_action_plans_group.py
```diff
@@ -0,0 +1,32 @@
...
+    group.permissions.add(view_action_plans_permission)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("user", "0023_create_barrier_search_download_approved_group"),
+    ]
+
+    operations = [migrations.RunPython(create_action_plans_group)]
```
kevincarrogan, line 32: Without specifying `reverse_code` this means that this migration is irreversible, is that what you meant for it to be?

api/user/serializers.py
```diff
@@ -105,6 +108,15 @@ def get_permitted_applications(self, obj):
             return sso_me.get("permitted_applications", None)
         return None
 
+    def get_has_approved_digital_trade_email(self, obj):
+        email_domain = obj.email.split("@")[-1].lower()
+        print(
+            "settings.APPROVED_DIGITAL_TRADE_EMAILS",
+            settings.APPROVED_DIGITAL_TRADE_EMAIL_DOMAINS,
+            email_domain,
+        )
```
kevincarrogan, line 117: Remove the debug or make a logging call.

kevincarrogan (changes requested):

api/user/migrations/0024_create_action_plans_group.py
jamespacileo: added reverse code for the migration

api/barriers/models.py
jamespacileo: face palm moment, fixed to being just an f-string

api/user/serializers.py
jamespacileo: removed

api/barriers/migrations/0114_auto_20211126_1422.py
```diff
@@ -0,0 +1,45 @@
...
+            ),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name="barriersearchcsvdownloadevent",
+            name="email",
+            field=models.EmailField(default="", max_length=254),
+            preserve_default=False,
+        ),
+        migrations.RunPython(remove_existing_download_entries),
```
kevincarrogan, line 44: This doesn't have a `reverse_code` so isn't reversible.

jamespacileo: added noop as reverse_code

kevincarrogan (approved):

PR #353, opened 2022-02-28
TSS-120 Adding Top Priority tag to DB

Santino-Trade:


Santino-Trade:
Had to change the unit tests as whenever the test was creating a new barriertag, it would keep trying to add it with an id of 4 - causing a fail because of the new 4th tag. Couldn't see why it was ignoring the new one.

jamespacileo (approved):

PR #368, opened 2022-04-04
TSS-24 - Moving complete % calculation from frontend and storing in DB

Santino-Trade:


NickFitz:
Bitten by the Black version problem 😦

NickFitz (approved):
👍

PR #369, opened 2022-04-07
TSS-177 UTF-8 characters in downloads

NickFitz:
May fix the CSV download problem of Excel failing to detect the file's UTF-8 encoding.

Also: fix the Dockerfile, and update pre-commit.


NickFitz:
Works as desired on UAT 😺

r4vi (approved):

PR #375, opened 2022-04-28
TSS-203 Fix for Barrier `modified_on` values

NickFitz:
This adds a management command to patch the `modified_on` fields of `Barrier` objects that were inadvertently modified by previous migrations. It uses a management command rather than a migration as fixes for previous migrations have been included just in case, and frankly this is a pretty loathsome hack that shouldn't be applied automatically, but run manually in the full and awful knowledge of just how dreadful it is 😉

### DIFF
--- a/api/barriers/management/commands/ugly_hack_fixing_modified_on_field.py
+++ b/api/barriers/management/commands/ugly_hack_fixing_modified_on_field.py
@@ -1,0 +1,19 @@
+from django.core.management import BaseCommand
+from django.db.models import Max, OuterRef, Subquery
+
+from ...models import Barrier
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


Santino-Trade (approved):

NickFitz:
> Not sure about calling it "ugly hack fix" in the code, but i'll roll with it 😅

Honesty is the best policy 😁 

> Wonder if theres a solution we can implement so it doesn't happen again? (besides us remembering to use 'update' instead of 'save' in a barrier migration)

I was thinking about that. One thought was to add another field like `user_modified_on` which isn't `auto_now` and gets explicitly set when the user actually does something. But that would involve a lot of changes, so this seems like the quickest and easiest thing for now, and we can have a think about that or other ways to dodge the problem in future.

PR #376, opened 2022-04-29
Updating Black version for local docker builds

Santino-Trade:


NickFitz (approved):
YES! Thank you 😁 

I kept meaning to add this in a commit but kept forgetting. No more opening a terminal on the Docker container and typing `pip install -U black` 😸

PR #388, opened 2022-05-23
Tss 208 prev estimation date csv column

Santino-Trade:


api/barriers/models.py
```diff
@@ -589,6 +589,13 @@ def is_resolved(self):
     def is_top_priority(self):
         return self.tags.filter(is_top_priority_tag=True).exists()
 
+    # @property
```
NickFitz, line 592: I assume these are here by accident? 🙂

NickFitz (approved):

PR #436, opened 2022-08-31
TSS-414

jamespacileo:


api/feedback/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.urls import path
```
NickFitz, line 1: We shouldn't need this - the DataFlow URL is defined in `api/dataset/urls.py` and any other views/URLs are out of scope.

NickFitz (changes requested):

api/feedback/migrations/0003_alter_feedback_created_on.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.13 on 2022-09-01 08:41
```
NickFitz, line 1: Could you not roll back the 0002 migration, delete it, and run `make migrations` again? Seems a bit tidier than having two migrations. As you're busy with the Hays stuff this morning, I can pull down your branch and do that if it's OK with you?

NickFitz (changes requested):

PR #446, opened 2022-09-16
TSS-397

jamespacileo:


api/barriers/serializers/data_workspace.py
```diff
@@ -265,7 +265,7 @@ def get_team_count(self, obj):
     def get_action_plan_added(self, obj):
         if not obj.action_plan:
             return False
-        return obj.action_plan.milestones.count() > 0
+        return obj.action_plan.get_active_action_plans.exists()
```
NickFitz, line 268: Should this be `get_active_action_plans()`?

PR #458, opened 2022-10-18
TSS-268

jamespacileo:


NickFitz (approved):
I wouldn't object to the various occurrences of `UserActvitiy` being corrected, but I suppose the runtime doesn't care about typos as long as they're consistent and it would be nice to get this shipped. So don't bother if it'll take too long 🙂

PR #463, opened 2022-10-26
Tss 483 new priorty form rebased onto 463

Santino-Trade:


api/barriers/serializers/base.py
```diff
@@ -61,6 +61,7 @@ class BarrierSerializerBase(
     last_seen_on = serializers.SerializerMethodField()
     modified_by = UserField(required=False)
     priority = BarrierPriorityField(required=False)
```
NickFitz, line 63: Looks like I missed this bit - does this need to go too?

api/barriers/serializers/csv.py
```diff
@@ -51,6 +51,7 @@ class BarrierCsvExportSerializer(AssessmentFieldsMixin, serializers.Serializer):
     product = serializers.CharField()
     source = serializers.SerializerMethodField()
     priority = serializers.SerializerMethodField()
```
NickFitz, line 53: Also missed this one…

api/barriers/serializers/data_workspace.py
```diff
@@ -201,6 +201,7 @@ class Meta(BarrierSerializerBase.Meta):
             "modified_on",
             "other_source",
             "priority",
```
NickFitz, line 203: And another! 😱

NickFitz (changes requested):

api/barriers/serializers/base.py
Santino-Trade: Just checked, we've got to keep it in the csv and DW, so think we need to keep this stuff

NickFitz: No, would be the correct answer 🤦‍♂️

NickFitz (approved):

PR #481, opened 2023-01-20
Modifying top_priority filter to reflect new search rule

Santino-Trade:


api/barriers/models.py
```diff
@@ -1357,6 +1357,13 @@ def ignore_all_sectors_filter(self, queryset, name, value):
 
     def top_priority_status_filter(self, queryset, name, value):
         if value:
+
+            # If user is searching for APPROVED top priority barriers, the search must also
+            # include barriers PENDING REMOVAL. So if APPROVED is selected, but PENDING
+            # REMOVAL has not, we need to include it in the search parameter.
+            if "APPROVED" in value and "REMOVAL_PENDING" not in value:
+                value.append("REMOVAL_PENDING")
```
ferozerub, line 1365: @Santino-Trade  - do we need to update the front end to reflect this i.e. when you click approved it also selects removal pending

ferozerub (approved):

api/barriers/models.py
Santino-Trade: Not this time, no, it's specifically listed as out of scope (think its why the help text was gone for instead)

PR #518, opened 2023-04-27
feature/TSS-558_dmas-add-dmas-data-to-workspace

osimuka:


Santino-Trade (approved):

api/barriers/serializers/data_workspace.py
```diff
@@ -246,6 +261,19 @@ class Meta(BarrierSerializerBase.Meta):
...
+            "programme_fund_progress_update_author",
+            "programme_fund_progress_update_date",
+            "programme_fund_progress_update_expenditure",
+            "programme_fund_progress_update_milestones",
+            "progress_update_author",
+            "progress_update_date",
+            "progress_update_message",
+            "progress_update_next_steps",
+            "progress_update_status",
+            "top_priority_summary",
```
ferozerub, line 276: Need to add NextSteps related objects

ferozerub (changes requested):

api/barriers/serializers/data_workspace.py
osimuka: added `next_steps_items` to the serialiser

api/barriers/serializers/csv.py
```diff
@@ -487,7 +487,7 @@ def get_previous_estimated_resolution_date(self, obj):
                 .exclude(estimated_resolution_date=obj.estimated_resolution_date)
                 .latest("history_date")
             )
-        except HistoricalBarrier.DoesNotExist:
+        except ObjectDoesNotExist:
```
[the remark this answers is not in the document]
osimuka: Yes it does not exist, thats why we are using the base `ObjectDoesNotExist` object which we imported and can account for. I believe `HistoricalBarrier` is created dynamically at runtime.

PR #544, opened 2023-07-18
feature/TSS-787/add-new-fields-from-mve-to-data-workspace

osimuka:


api/barriers/serializers/data_workspace.py
```diff
@@ -184,6 +185,9 @@ class DataWorkspaceSerializer(AssessmentFieldsMixin, BarrierSerializerBase):
     proposed_estimated_resolution_date = serializers.SerializerMethodField()
     proposed_estimated_resolution_date_user = serializers.SerializerMethodField()
     proposed_estimated_resolution_date_created = serializers.SerializerMethodField()
+    main_sector = serializers.SerializerMethodField()
```
chris-pettinga, line 188: Can we not use the SectorsField that Feroze created

chris-pettinga (changes requested):

api/barriers/serializers/data_workspace.py
osimuka: Yes, but I will be looking for the string name itself, I have adjusted the method to use `SectorField` thanks for pointing it out.
The dataworkspace is looking for a string or None in that field

chris-pettinga: got you, didn't realise this was specifically the DataWorkspaceSerializer

osimuka: No worries, but the pointers helped very well thanks

api/barriers/serializers/data_workspace.py
```diff
@@ -478,3 +488,22 @@ def get_proposed_estimated_resolution_date_user(self, instance):
...
         )
         return f"{first_name} {last_name}" if first_name and last_name else None
+
+    def get_main_sector(self, instance) -> typing.Optional[str]:
+        main_sector = SectorField(instance=instance, required=False)
+        if main_sector.is_valid():
+            return main_sector.data["name"]
+        return None
+
+    def get_export_types(self, instance) -> typing.List[str]:
```
chris-pettinga, line 498: don't need this method if export_types is now a ExportTypes field

ferozerub (approved):

chris-pettinga (approved):

PR #545, opened 2023-07-21
feature/TSS-859/add-export-description-field-to-dataworkspace

osimuka:


api/barriers/fields.py
```diff
@@ -466,3 +466,13 @@ def to_internal_value(self, data):
             if val == data:
                 return key
         self.fail("invalid_choice", input=data)
+
+
+class LineBreakCharField(serializers.CharField):
```
chris-pettinga, line 471: i like this

PR #552, opened 2023-07-31
feature/TSS-882/add-new-report-a-barrier-fields-to-csv

osimuka:


chris-pettinga (changes requested):
If both get_export_types() methods on the serializers are the same, I think it makes sense to have a new ExportTypeNames field that has that logic in the to_representation() method?

chris-pettinga (approved):
Looks good, thank you.

PR #561, opened 2023-08-07
feature/TSS-889/add-new-search-fields-included-and-changedfor-mve

osimuka:


api/barriers/models.py
```diff
@@ -1390,7 +1393,9 @@ def sector_filter(self, queryset, name, value):
         custom filter for multi-select filtering of Sectors field,
         which is ArrayField
         """
-        return queryset.filter(Q(all_sectors=True) | Q(sectors__overlap=value))
+        return queryset.filter(Q(main_sector__in=value)).filter(
+            Q(all_sectors=True) | Q(sectors__overlap=value)
+        )
```
Santino-Trade, line 1398: We'll need a unit test for this - This seems to be one of those Ticket vs Design situations - from what i understand, we need a tickbox to indicate that we only want the search term to apply to main_sector, or if we want the search term to apply for main_sector AND sector - is that the behaviour this filter is going to replicate?

osimuka: @Santino-Trade I had a discussion with @ferozerub prior, that field is no longer required

tests/barriers/test_list_barriers.py
```diff
@@ -499,6 +499,45 @@ def test_list_barriers_text_filter_based_on_summary(self):
...
+        barrier.save()
+
+        barrier2 = BarrierFactory()
+
+        barrier2.companies = [
+            {"id": "10", "name": "Springfield Nuclear Power Plant LTD"},
+        ]
+        barrier2.save()
+
+        url = f'{reverse("list-barriers")}?search=test company 3'
```
Santino-Trade, line 534: Linked to the comment above, can we do a unit test here to test for the "Agent" company, to show the search applies for any company in a longer list of companies

api/barriers/models.py
```diff
@@ -1553,11 +1558,15 @@ def text_search(self, queryset, name, value):
             full text search on summary
             partial search on title
         """
-        return queryset.annotate(search=SearchVector("summary"),).filter(
+        return queryset.annotate(
+            search=SearchVector("summary", "export_description"),
+            company_name=RawSQL("(companies->0->>'name')", ()),
```
Santino-Trade, line 1563: Next catchup, can you walk us through "(companies->0->>'name')" ? is this looping every company in the companies column? Turning the column into a single string?

ferozerub (approved):

PR #568, opened 2023-08-21
TSS 931 remove duplicates in search results

ferozerub:


api/barriers/views.py
```diff
@@ -263,56 +261,6 @@ def perform_update(self, serializer):
...
-        # then append those that don't, ordered by the default ordering.
-        ordering = request.query_params.get(self.ordering_param)
-        ordering_config = BARRIER_SEARCH_ORDERING_CHOICES.get(ordering, None)
-        if ordering_config is None:
-            # Not one of our fancy ones, so just do the usual
-            return super().filter_queryset(request, queryset, view)
-        # Need to bear in mind that chaining querysets can cause performance problems
-        # if it results in them being evaluated early,
-        # so we go to great lengths to ensure the real work is handed off to the DB.
-        partition_on = ordering_config["ordering-filter"]
```
Santino-Trade, line 278: We need to add the ordering filter somewhere in the new implementation - ordering "resolved" looks at status_date, and without the filter saying "ONLY TAKE RESOLVED:IN-FULL" we'll be seeing unresolved barriers in the wrong order.

Santino-Trade: Yea, i'm trying now to find a way of NOT doing this split method while trying to ensure a "resolved date" filter works. It would need;
Resolved In Full barriers first, sorted by their status-date
All other barriers, sorted by their reported-on date

chris-pettinga (approved):

PR #579, opened 2023-09-11
fix/TSS-968/DMAS-main-sector-to-be-added-in-search-filter

osimuka:


tests/barriers/test_list_barriers.py
```diff
@@ -381,6 +381,18 @@ def test_list_barriers_sector_filter(self):
...
+    def test_list_barriers_only_main_sector_filter(self):
+        barrier = BarrierFactory()
+
+        assert Barrier.objects.count() == 1
+
+        url = f'{reverse("list-barriers")}?sector={barrier.main_sector}&only_main_sector=yes'
+        response = self.api_client.get(url)
+
+        assert status.HTTP_200_OK == response.status_code
+        assert response.data["count"] == 1
```
chris-pettinga, line 393: can we add a new test confirming that if we make a request with another sector ID (that is not the barriers main sector), the API returns 0 results?

chris-pettinga (changes requested):

tests/barriers/test_list_barriers.py
osimuka: Nice, added the new test case

api/barriers/models.py
```diff
@@ -1395,9 +1395,23 @@ def sector_filter(self, queryset, name, value: List[str]):
...
+            return queryset.filter(main_sector__isnull=False).distinct()
+
+        # Define base query based on provided options
+        base_query = Q(all_sectors=not ignore_all_sectors) | Q(main_sector__in=value)
+
+        if only_main_sector == "yes":
+            return queryset.filter(base_query).distinct()
+
+        # Add overlap condition for sectors
+        return queryset.filter(base_query | Q(sectors__overlap=value)).distinct()
```
chris-pettinga, line 1414: I think we can remove the .distinct() call here as Feroze's ticket should've resolved these duplicates
```suggestion
        return queryset.filter(base_query | Q(sectors__overlap=value))
```

api/barriers/models.py
```diff
@@ -1395,9 +1395,23 @@ def sector_filter(self, queryset, name, value: List[str]):
...
+        # If we're only concerned with main sectors and there's no provided value,
+        # we can directly check if main_sector is not null
+        if only_main_sector == "yes" and not value:
+            return queryset.filter(main_sector__isnull=False).distinct()
+
+        # Define base query based on provided options
+        base_query = Q(all_sectors=not ignore_all_sectors) | Q(main_sector__in=value)
+
+        if only_main_sector == "yes":
+            return queryset.filter(base_query).distinct()
```
chris-pettinga, line 1411: Same as here
```suggestion
            return queryset.filter(base_query)
```

api/barriers/models.py
```diff
@@ -1395,9 +1395,23 @@ def sector_filter(self, queryset, name, value: List[str]):
         custom filter for multi-select filtering of Sectors field,
         which is ArrayField
         """
-        return queryset.filter(
-            Q(all_sectors=True) | Q(main_sector__in=value) | Q(sectors__overlap=value)
-        ).distinct()
+
+        ignore_all_sectors = self.data.get("ignore_all_sectors", False)
+        only_main_sector = self.data.get("only_main_sector", "no")
```
chris-pettinga, line 1400: for consistency, I think this should also default to False, then we don't need to check for `only_main_sector == "yes"`, we can just do `if only_main_sector`

api/barriers/models.py
```diff
@@ -1395,9 +1395,23 @@ def sector_filter(self, queryset, name, value: List[str]):
...
+
+        # If we're only concerned with main sectors and there's no provided value,
+        # we can directly check if main_sector is not null
+        if only_main_sector == "yes" and not value:
+            return queryset.filter(main_sector__isnull=False).distinct()
```
chris-pettinga, line 1405: remove .distinct()
```suggestion
            return queryset.filter(main_sector__isnull=False)
```

chris-pettinga (changes requested):

chris-pettinga (approved):

ferozerub (approved):

PR #581, opened 2023-09-11
Feature - TSS-900 - Added user context to sentry

chris-pettinga:


osimuka (approved):

api/core/middleware.py
```diff
@@ -1,31 +1,25 @@
...
-                    remote_address = (
-                        request.META["HTTP_X_FORWARDED_FOR"].split(",")[-2].strip()
-                    )  # noqa: E501
-                except (IndexError, KeyError):
-                    logger.warning(
-                        "X-Forwarded-For header is missing or does not "
-                        "contain enough elements to determine the "
-                        "client's ip"
-                    )
-                    return HttpResponse("Unauthorized", status=401)
```
ferozerub, line 24: I think this restricts access to django admin in production. I am not sure we can remove it completely.

chris-pettinga: @ferozerub - The code scythe knows no remorse, it is not referenced anywhere in the codebase or in Vault

PR #588, opened 2023-09-21
pre-commit fixes

osimuka:


api/barriers/serializers/base.py
```diff
@@ -38,15 +38,49 @@
...
 from api.metadata.fields import AdminAreasField, CountryField, TradingBlocField
+from api.metadata.constants import ECONOMIC_ASSESSMENT_IMPACT
 
 from .mixins import LocationFieldMixin
 from .public_barriers import NestedPublicBarrierSerializer
 
 logger = logging.getLogger(__name__)
 
 
+class BarrierMixins(metaclass=serializers.SerializerMetaclass):
```
abarolo, line 49: Should this be renamed to something a bit less generic? BarrierSerializerMethodMixins or something? Just thinking BarrierMixins name suggests it has core elements for Barriers.

abarolo (commented):
Do we need to start adding unit tests to reflect changes like this in API?

osimuka:
> Do we need to start adding unit tests to reflect changes like this in API?

There is tests already put in place, this is to optimise the serializer as the target fields in the tests have already been tested!

abarolo:
Yeah I see they are tested in the list_barriers*.py, but theres no explicit tests for the full response data object being returned from the serializer. Perhaps thats something we should have / worth a discussion with team about? As it is now it's easy to add and remove fields from a serializer without it affecting tests.

abarolo (approved):

api/barriers/serializers/base.py
osimuka: Good point the name of the serializer is now `BarrierBaseMixins` as its required by different barrier `read` serializers

ferozerub (approved):

PR #602, opened 2023-10-04
Adding 'archived_on' back to barrier list serializer

Santino-Trade:


chris-pettinga (commented):
Can we add a test for this serializer to make sure fields are not accidentally removed again in the future

ferozerub (approved):

PR #633, opened 2023-11-01
Feature - TSS-1081 - DW Pipeline main_sector change

chris-pettinga:
Main Sector” should show “all sectors” IF, “main sector” is null AND "All Sectors” is TRUE.

Added tests

### DIFF
--- a/api/barriers/serializers/data_workspace.py
+++ b/api/barriers/serializers/data_workspace.py
@@ -501,8 +501,8 @@ class DataWorkspaceSerializer(AssessmentFieldsMixin, BarrierSerializerBase):
         if main_sector:
             return main_sector["name"]
         elif instance.all_sectors:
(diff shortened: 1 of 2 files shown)


api/barriers/serializers/data_workspace.py
```diff
@@ -500,6 +500,10 @@ def get_main_sector(self, instance) -> typing.Optional[str]:
         main_sector = metadata_utils.get_sector(instance.main_sector)
         if main_sector:
             return main_sector["name"]
+        elif instance.all_sectors:
+            # if there is no main sector, but there are all sectors, return "all sectors"
+            return "all sectors"
```
ferozerub, line 505: Perhaps we should use 'All sectors' as it is represented on the front end.

osimuka (approved):

PR #634, opened 2023-11-01
TSS-982: Add Priority Level to history

abarolo:


api/history/items/barriers.py
```diff
@@ -7,6 +7,7 @@
 
 from ...barriers.models import BarrierTopPrioritySummary
 from .base import BaseHistoryItem
+from ...metadata.constants import PRIORITY_LEVELS
```
chris-pettinga, line 10: I know the relative ... imports have been around since before this PR, but I think they're ugly and importantly liable to fail. Would you mind refactoring them so they're more explicit? Thanks

api/history/items/barriers.py
```diff
@@ -251,3 +252,10 @@ def get_value(self, record):
 class TopPrioritySummaryHistoryItem(BaseHistoryItem):
     model = "barrier_top_priority_summary"
     field = "top_priority_summary_text"
+
+
+class PriorityLevelHistoryItem(BaseBarrierHistoryItem):
+    field = "priority_level"
+
+    def get_value(self, record):
+        return PRIORITY_LEVELS._display_map.get(record.priority_level, "")
```
chris-pettinga, line 261: can you use record.get_priority_level_display() here instead?

```suggestion
        return record.get_priority_level_display()
```

chris-pettinga (changes requested):

chris-pettinga (approved):

ferozerub (approved):

osimuka (approved):

PR #635, opened 2023-11-01
Creating new group with permission to edit users

Santino-Trade:


api/user/migrations/0032_role_admin_permissions.py
```diff
@@ -0,0 +1,32 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("user", "0031_alter_useractvitiylog_event_type"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            assign_permissions, reverse_code=migrations.RunPython.noop
```
chris-pettinga, line 30: can we add the reverse function in case we do actually want to rollback?

chris-pettinga (changes requested):

chris-pettinga (approved):

PR #638, opened 2023-11-03
Feature/tss-1064-display-related-barriers

osimuka:


abarolo:
Can we add some unit tests to test the view works and some to test `get_similar_barriers`?

osimuka:
> Can we add some unit tests to test the view works and some to test `get_similar_barriers`?

Yes we can add tests, since its an ML functionality, the main tests will be mocked data and making sure functions are being called.

PR #640, opened 2023-11-07
Feature/tss 1093/review access control

ferozerub:


tests/feedback/test_views.py
```diff
@@ -11,7 +16,14 @@
...
 
 
-class MyTestCase(APITestMixin, APITestCase):
+class FeedbackTestCase(APITestMixin, APITestCase):
+    def setUp(self):
+        super().setUp()
+        self.user1 = create_test_user(sso_user_id=self.sso_user_data_1["user_id"])
+        self.user2 = create_test_user(sso_user_id=self.sso_user_data_2["user_id"])
+        self.api_client1 = self.create_api_client(user=self.user1)
+        self.api_client2 = self.create_api_client(user=self.user2)
```
chris-pettinga, line 25: Looks like we don't use this further on down?

chris-pettinga (changes requested):

chris-pettinga (approved):

tests/feedback/test_views.py
ferozerub: Agreed - removed the second api client

api/core/permissions.py
```diff
@@ -8,7 +8,23 @@ class IsAuthenticated(BasePermission):
...
+class IsCreatorOrReadOnly(BasePermission):
+    """
+    Object-level permission to only allow creators of an object to edit it.
+    Assumes the model instance has an `created_by` attribute.
+    """
+
+    def has_object_permission(self, request, view, obj):
+        # Read permissions are allowed to any request,
+        # so we'll always allow GET, HEAD or OPTIONS requests.
+        if request.method in SAFE_METHODS:
```
osimuka, line 26: if this is a get request, in DRF which is also `retrieve` & `list` it will return True

Santino-Trade (approved):

api/core/permissions.py
osimuka: is this what we want here?

osimuka (approved):

PR #647, opened 2023-11-19
TSS-958: History V2 POC

abarolo:
The barrier history endpoint scales inefficiently.

Legacy query complexity: ~O(N x M  x R) 
 
N= # historical tables
M = # items
R = # item related lookups

Below is a screenshot of a legacy barrier history tab with ~20 historical items on separate tables.

Screenshot 2023-11-18 at 23 45 19

This is a POC of working functionality backwards compatible with the remaining non-refactored legacy implementation.

Once all historical components are fully refactored:

V2 query complexity = O(N)

N = # historical tables

TODO: 
- add get_history() to all tracked models.
- delete all historical factory and item tech debt ~2000 lines + ~20 files
- certain historical data items may need enrichment, so a solution to that.


ferozerub (approved):

tests/history/factories.py
```diff
@@ -0,0 +1,31 @@
+import factory
+
+from api.barriers.models import Barrier, ProgrammeFundProgressUpdate
+
+
+class BarrierFactory(factory.django.DjangoModelFactory):
```
chris-pettinga, line 6: We already have a BarrierFactory here - https://github.com/uktrade/market-access-api/blob/588198500a067b78230b492de3e4425f7f8163ca/tests/barriers/factories.py#L78

chris-pettinga (changes requested):

tests/history/factories.py
abarolo: Yes I got it from here and removed parts of it. Just trying to maintain as much separation for now while doing the refactor.

chris-pettinga: ok happy with this as long as you keep track of all the additional code infrastructure the refactor needs, so we're able to quickly delete all of it/all of the old stuff when we do the absolute switchover.

chris-pettinga (approved):

tests/history/factories.py
abarolo: of coourse

PR #648, opened 2023-11-20
Release - Man Uka of Lord - a

ferozerub:
* adding pb100 to priority level

* changed priority level to a method field

### DIFF
--- a/api/barriers/serializers/data_workspace.py
+++ b/api/barriers/serializers/data_workspace.py
@@ -191,6 +191,7 @@ class DataWorkspaceSerializer(AssessmentFieldsMixin, BarrierSerializerBase):
     trade_direction = serializers.SerializerMethodField()
     export_description = LineBreakCharField(required=False)
     tags = serializers.SerializerMethodField()
+    priority_level = serializers.SerializerMethodField()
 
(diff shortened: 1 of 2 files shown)


api/barriers/serializers/data_workspace.py
```diff
@@ -513,3 +514,11 @@ def get_trade_direction(self, instance) -> typing.Optional[str]:
             )
         else:
             return None
+
+    def get_priority_level(self, instance) -> typing.Optional[str]:
+        if (
+            instance.priority_level == "NONE"
+            and instance.top_priority_status == "APPROVED"
```
chris-pettinga, line 521: this should also include "REMOVAL_PENDING" and "RESOLVED" I think

```suggestion
            and instance.top_priority_status in ("APPROVED", "REMOVAL_PENDING", "RESOLVED")
```

chris-pettinga (changes requested):

api/barriers/serializers/data_workspace.py
osimuka: the ticket just said "APPROVED" state for top priority

osimuka (approved):

PR #659, opened 2023-11-30
TSS-1152: BarrierFactory history endpoint now uses V2

abarolo:


chris-pettinga:
The enrichment is confusing me a bit, is it just legacy functionality that allows it to work with V1 code?

abarolo:
> The enrichment is confusing me a bit, is it just legacy functionality that allows it to work with V1 code?

The legacy implementation would, in some cases, change the raw changed data at the Item level for certain fields.

IE:

`
class LocationHistoryItem(BaseBarrierHistoryItem):
    field = "location"

    def get_value(self, record):
        return get_location_text(
            country_id=record.country,
            trading_bloc=record.trading_bloc,
            caused_by_trading_bloc=record.caused_by_trading_bloc,
            admin_area_ids=record.admin_areas,
        )
`

V2 does everything in steps kind of like a pipeline:

`
1) Get the raw changes for a model
2) Enrich where needed (As required by the UI (backwards compatibility in this case and we don't need to update any FE))
`

This gives us some flexibility, rather than coupling everything together at the lowest level (A legacy history Item). Also, some history items will be making DB calls, so with this pipeline approach we can make 1 db call to get everything, rather than a DB call for each item.

chris-pettinga (approved):

osimuka (approved):

osimuka:
@abarolo nice work 💯

PR #673, opened 2023-12-05
Merged Top 100 and priority search

ferozerub:


api/barriers/models.py
```diff
@@ -1473,6 +1474,30 @@ def priority_filter(self, queryset, name, value):
...
+            # If user is searching for APPROVED top priority barriers, the search must also
+            # include barriers PENDING REMOVAL. So if APPROVED is selected, but PENDING
+            # REMOVAL has not, we need to include it in the search parameter.
+            if "APPROVED" in value and "REMOVAL_PENDING" not in value:
+                value.append("REMOVAL_PENDING")
+
+            queryset = queryset.filter(
+                Q(top_priority_status__in=value) | Q(priority_level__in=value)
+            )
+            if "NONE" in value:
```
chris-pettinga, line 1493: if "NONE" is in the value, then doesn't the above filter do the exact same thing as the filter below? e.g `value = ["NONE"]` then `Q(top_priority_status__in=value)` is the exact same as `Q(top_priority_status="NONE")`

chris-pettinga (changes requested):
Looks good but needs tests as coverage is decreased.

osimuka (approved):

chris-pettinga (approved):

PR #677, opened 2023-12-08
TSS-1364: Upgrade barrier history to use V2

abarolo:


api/history/manager.py
```diff
@@ -107,18 +135,15 @@ def get_full_history(cls, barrier, ignore_creation_items=False, use_cache=False)
...
-        history += cls.get_economic_impact_assessment_history(
-            barrier.pk, start_date=start_date
-        )
-        history += cls.get_resolvability_assessment_history(
-            barrier.pk, start_date=start_date
-        )
-        history += cls.get_strategic_assessment_history(
-            barrier.pk, start_date=start_date
-        )
+        # history += cls.get_economic_impact_assessment_history(
```
Santino-Trade, line 138: remove old stuff

api/assessment/models.py
```diff
@@ -114,6 +116,30 @@ def save(self, *args, **kwargs):
...
+        default_fields = (
+            "approved",
+            "archived",
+            "documents_cache",
+            "explanation",
+            "export_value",
+            "import_market_size",
+            "rating",  # Needs enrichment
+            "ready_for_approval",
+            "value_to_economy",
```
Santino-Trade, line 131: Do we need all these fields? Are they all sucked up by the frontend and not used, or do we display all of these fields on the history cards? (Same question for the other models)

api/assessment/models.py
```diff
@@ -114,6 +116,30 @@ def save(self, *args, **kwargs):
...
+    def get_history(cls, barrier_id: str, fields: Optional[List] = None):
+        qs = cls.history.filter(barrier__id=barrier_id)
+        default_fields = (
+            "approved",
+            "archived",
+            "documents_cache",
+            "explanation",
+            "export_value",
+            "import_market_size",
+            "rating",  # Needs enrichment
```
Santino-Trade, line 129: remove enrichment reminders

api/assessment/models.py
```diff
@@ -1,3 +1,4 @@
+from typing import List, Optional
```
Santino-Trade, line 1: Easier to see with a comment here, but theres a failing U test

abarolo: That test will be redundant https://github.com/uktrade/market-access-api/pull/676

PR #679, opened 2023-12-08
OLD - Part 1 of TSS-1363 

Santino-Trade:


api/history/v2/enrichment.py
```diff
@@ -195,3 +197,71 @@ def enrich(value):
...
+        owner_result = User.objects.filter(id=value)
+        for owner in owner_result:
+            full_name = owner.first_name + " " + owner.last_name
+            return full_name
+
+        # If there was no result, there was no ID the
+        # action plan was previously owned by or is going to be owned by
+        return None
+
+    cleaned_history = []
```
abarolo, line 213: Is this necessary? Can't you just update the values from teh original dict.

Also, the `strategic_context`, does it need to be handled at all? The legacy history just displays the changes with no enrichment
`

Santino-Trade: I can do it there in the manager, i can do it here in enrichment. Personally, i think we should clean any values in one place, not more, and I can't seen any other fields being done elsewhere. As for strategic context, we need to do something or as stated in the comment, empty history items show up. BTW, this work isn't ready for review.

Santino-Trade: With your pointer about lists the other day, i've been able to source the name parts from the initial query, so won't have to make a second to the User table 👍

PR #701, opened 2024-01-12
TSS-1378: Barrier update history + next steps

abarolo:


osimuka (approved):

tests/history/factories.py
```diff
@@ -38,3 +43,15 @@ class ProgrammeFundProgressUpdateFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = ProgrammeFundProgressUpdate
+
+
```
Santino-Trade, line 47: Are the tests that use the factories complex enough to warrant their use? It'd be simpler to just create the respective test objects using the model directly. I can't remember if it populates extra mandatory fields automatically though or if you have to do that here or when instantiating the factory in the test - if you're passing the important stuff you might aswell cut out the middleman.

abarolo: I suppose we could use the models directly. imo i'd rather create a factory for all the models where possible for consistency and general testing practice, rather than using factory on some models?

Santino-Trade (approved):

PR #704, opened 2024-01-16
adding main sector to sectors for public view and public barriers

osimuka:


abarolo:
Can you add test to reflect internal_main_sector being added?

api/barriers/serializers/public_barriers.py
```diff
@@ -314,8 +319,10 @@ def get_sectors(self, obj):
         if obj.all_sectors:
             return [{"name": "All sectors"}]
         else:
+            # we need to add the main sector into the list as the first field of the array
+            sectors = [obj.internal_main_sector] + obj.sectors
```
Santino-Trade, line 323: If we're handing the frontend main_sector as a separate field in line 236, do we need to combine them together in the regular sector field? It looks like the change on python-frontend repo combines them into a list - wouldn't this result in the main sector popping up twice?

Santino-Trade (approved):

api/barriers/fields.py
```diff
@@ -409,6 +409,10 @@ class ReadOnlySectorsField(FilterableReadOnlyField):
     Field serializer to be used with read only sectors fields.
     """
 
+    def __init__(self, to_repr_keys=(), no_sort=False, **kwargs):
```
abarolo, line 412: can it be sort=True instead rather than double negative.

abarolo (approved):

api/barriers/fields.py
osimuka: good idea, makes it more readable

PR #713, opened 2024-01-24
MAU RELEASE BRANCH

Santino-Trade:
…712)


api/barriers/views.py
```diff
@@ -900,18 +900,63 @@ def update_status_action(self, public_view_status):
...
+    @action(methods=["post"], detail=True, permission_classes=(IsApprover,))
     def unprepared(self, request, *args, **kwargs):
-        return self.update_status_action(PublicBarrierStatus.ELIGIBLE)
+        return self.update_status_action(PublicBarrierStatus.ALLOWED)
 
     @action(
         methods=["post"],
         detail=True,
-        permission_classes=(IsEditor,),
+        permission_classes=(),
```
Santino-Trade, line 916: UPDATE PUBLISH METHOD:

extra protection - check that the barrier is "AWAITING_PUBLISHING" before moving to "PUBLISHED" - to avoid being able to jump from other statuses

osimuka (approved):

api/barriers/models.py
```diff
@@ -1146,12 +1152,10 @@ def public_view_status(self):
```
Santino-Trade, line 1152: here

PR #727, opened 2024-02-25
TSS-158: Related Barrier MVP

abarolo:


api/related_barriers/manager.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+        .values("id", "barrier_corpus")
+    )
+
+
+def init():
+    global manager
+
+    if manager:
+        raise Exception("DB already set, please stop db or restart application")
```
osimuka, line 213: Is this for the DB of the application or for the models?

tests/related_barriers/test_manager.py
osimuka, line 1: nice tests 💯

api/related_barriers/manager.py
abarolo: Nice catch, will update the name

osimuka (approved):

ferozerub (approved):

PR #742, opened 2024-03-15
hotfix/tss-1625: Move torch to requirement.txt

abarolo:


requirements.txt
osimuka, line 1: @abarolo so the current `requirement.txt` file is updated using poetry (so the change will be overwritten)
to have this type of setup I suggest you create a new requirements file `requirements-ml.txt`
which will import `requirement.txt` and have the lines:
```
# DO NOT DELETE: (Related Barriers)
sentence-transformers==2.2.2
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.0.0
torchvision==0.15.1
```
then we point the build to use the `requirements-ml.txt`
@Santino-Trade @ferozerub any opinions?

ferozerub (approved):

osimuka (approved):

PR #780, opened 2024-04-19
Removing history purge from data anonymise

Santino-Trade:


api/barriers/management/commands/create_mock_barriers.py
```diff
@@ -15,7 +15,7 @@
 
 
 class Command(BaseCommand):
-    help = "Publish fake barriers"
```
Santino-Trade, line 18: It doesn't publish them

abarolo (approved):

ferozerub (approved):

PR #794, opened 2024-05-20
feature/TSS-1790-move-notification-to-signals-to-celery-background-tasks

osimuka:


EPedley (approved):

abarolo (commented):
Looks good, but can we put all the tasks in the main tasks file `api.barriers.tasks`?

api/barriers/signals/handlers.py
```diff
@@ -39,24 +42,25 @@ def barrier_categories_changed(sender, instance, action, **kwargs):
...
-        if hasattr(instance, "categories_history_saved"):
-            historical_instance = HistoricalBarrier.objects.filter(
-                id=instance.pk
-            ).latest("history_date")
-            historical_instance.update_categories()
-            historical_instance.save()
-        else:
-            instance.categories_history_saved = True
-            instance.save()
+        with transaction.atomic():
```
abarolo, line 45: Do we need to use atomic?

osimuka: added it in so that all the transactions can be performed before committing, as there is update, read operations being performed

osimuka:
> Looks good, but can we put all the tasks in the main tasks file `api.barriers.tasks`?

Good Idea, all moved over to `api.barriers.tasks`

api/barriers/tasks.py
```diff
@@ -292,3 +292,92 @@ def send_barrier_inactivity_reminders():
...
+
+
+@shared_task
+def send_new_valuation_notification(barrier_id: int):
+    """
+    Create the email client and send the new valuation notification email
+    """
+
+    # avoid circular import
+    from api.barriers.models import Barrier
```
abarolo, line 304: You can remove this now

osimuka: good spotting

PR #812, opened 2024-06-19
TSS/1522: adding market distorting practices tag

EPedley:


ferozerub (approved):

api/metadata/migrations/0047_market_distorting_practices_tag.py
```diff
@@ -0,0 +1,39 @@
...
+    "title": "Market Distorting Practices (MDP)",
+    "description": "",
+    "show_at_reporting": True,
+}
+
+
+def create_market_distorting_practices_tag(apps, schema_editor):
+    BarrierTag = apps.get_model("metadata", "BarrierTag")
+
+    order = BarrierTag.objects.order_by("-order")[:1].get().order
```
abarolo, line 15: Just some food for thought:

Django will by default return all values of a model, so even though you only want `order`, this will actually return every field for that BarrierTag and evaluate in memory. So, in the network communication between the application and the db, you are passing more data than required, but also loading more data into the application memory.

Of course in this scenario, it doesn't really matter. But imagine if BarrierTag had 1000 columns, or if this was being run 1000s of times a day, thats a lot of extra overhead!

abarolo (approved):

api/metadata/migrations/0047_market_distorting_practices_tag.py
EPedley: Ah thanks, I hadn't thought of that! 

I used this because I wasn't sure what the "order" should be because the previous tag addition figured it out dynamically like this. But I think I'll change this so that "order" is hardcoded - if we want to create more tags in future, we'll either have to figure it out anyway or use this code again (I think) to figure it out dynamically. So it may be best to nip it in the bud and figure out what it should be now.

Thanks!

Santino-Trade (approved):

abarolo (approved):

PR #816, opened 2024-06-24
Removing code no longer used after implementing MAU

Santino-Trade:


tests/barriers/test_public_barriers.py
```diff
@@ -1148,8 +852,8 @@ def test_ready_to_be_published_is_true_for_republish(self):
         pb, response = self.publish_barrier(user=user)
         assert status.HTTP_200_OK == response.status_code
 
-        assert False is pb.unpublished_changes
-        assert False is pb.ready_to_be_published
+        assert pb.unpublished_changes == []
```
abarolo, line 855: pep guidelines would recommend:

```
assert not pb.unpublished_changes
assert not pb.read_to_be_published
```

api/barriers/models.py
```diff
@@ -835,7 +834,27 @@ def save(
         super().save(force_insert, force_update, using, update_fields)
 
         # Ensure that a PublicBarrier for this Barrier exists
-        PublicBarrier.public_barriers.get_or_create_for_barrier(barrier=self)
+        # Update its non-editable fields to match any updated values
+        public_barrier_instance = (
```
abarolo, line 838: `get_or_create_for_barrier` returns `Tuple[PublicBarrier, bool]` so `public_barrier_instance` is a slightly misleading name.

maybe:

`public_barrier, _ = PublicBarrier.public_barriers.get_or_create_for_barrier(barrier=self)` then you do't need to access it below with `[0]`

api/barriers/models.py
```diff
@@ -1174,157 +1083,57 @@ def public_view_status(self, value):
...
             or self.public_view_status == PublicBarrierStatus.UNPUBLISHED
         )
-        is_republish = self.unpublished_on is not None
-        has_changes = self.unpublished_changes
-        has_title_and_summary = bool(self.title and self.summary)
-        return is_ready and has_title_and_summary and (is_republish or has_changes)
+        return is_ready
 
     @property
     def unpublished_changes(self):
```
abarolo, line 1099: This is a property (read-only) so it as an anti-pattern for it to cause changes to the object? Perhaps introduce a function to do the changes?

api/barriers/models.py
```diff
@@ -1174,157 +1083,57 @@ def public_view_status(self, value):
...
+                            latest_version, field
+                        ):
+                            changed_list.append(field)
+                    else:
+                        if set(self.barrier.categories.all()) != set(
+                            latest_version.categories.all()
+                        ):
+                            changed_list.append(field)
+
+        if changed_list != []:
```
abarolo, line 1126: pep: `if changed_list:`

abarolo: And for the conditions below

api/barriers/models.py
```diff
@@ -835,7 +834,27 @@ def save(
         super().save(force_insert, force_update, using, update_fields)
 
         # Ensure that a PublicBarrier for this Barrier exists
-        PublicBarrier.public_barriers.get_or_create_for_barrier(barrier=self)
+        # Update its non-editable fields to match any updated values
+        public_barrier_instance = (
+            PublicBarrier.public_barriers.get_or_create_for_barrier(barrier=self)
+        )
+        non_editable_public_fields = [
```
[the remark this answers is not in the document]
Santino-Trade: They're not editable on the public barrier directly, they're only editable on the Barrier object but we need them on the public barrier so they get taken over to the published version on CITB

api/barriers/models.py
Santino-Trade: Yea, I wasn't too sure about this. Think i might split to a separate function to do the save - since we know it'd only be called on the barrier save operation and we won't have to skip past the save logic when the property is accessed elsewhere.

ferozerub (approved):

abarolo (approved):

PR #823, opened 2024-06-28
TSS-1865-dmas-filtering-issue

osimuka:


ferozerub (approved):

EPedley (approved):

pyproject.toml
```diff
@@ -44,7 +44,7 @@ sentry-sdk = "1.14.0"
 URLObject = "==2.4.3"
```
Santino-Trade, line 44: Did you need to do this upgrade for this change to work? it'll conflict with any other version bump PRs going into the release

tests/barriers/test_list_barriers.py
```diff
@@ -1219,6 +1220,14 @@ def test_additional_ordering_filters(self):
         assert response.data["results"][2]["id"] == str(barrier_4.id)
         assert response.data["results"][3]["id"] == str(barrier_3.id)
 
+    def test_valuation_assessment_filter(self):
```
abarolo, line 1223: Can you add a test for the wihtout?

PR #845, opened 2024-07-16
Adding free-text-search to related barriers

Santino-Trade:
New endpoint:

`/related-barriers?search_term=xxx`

Allows textual search of barrier embeddings


api/related_barriers/urls.py
```diff
@@ -12,4 +12,7 @@
     path(
         "barriers/<uuid:pk>/related-barriers", related_barriers, name="related-barriers"
     ),
+    path(
+        "barriers/dashboard/<str:search_term>", related_barriers_search, name="related-barriers-search"
```
abarolo, line 16: Perhaps `path("related-barriers", ...)` with the `search_term` as query param?

PR #874, opened 2024-08-27
TSS-2081: Policy Teams History Backend

EPedley:


api/barriers/models.py
```diff
@@ -138,26 +138,37 @@ class BarrierHistoricalModel(models.Model):
...
-            changed_fields.add("categories")
-
-        commodity_codes = [c.get("code") for c in self.commodities_cache]
-        old_commodity_codes = [c.get("code") for c in old_history.commodities_cache]
-        if set(commodity_codes) != set(old_commodity_codes):
-            changed_fields.add("commodities")
-
-        if set(self.tags_cache or []) != set(old_history.tags_cache or []):
-            changed_fields.add("tags")
+    def get_cached_fields(self, old_history, changed_fields):
```
abarolo, line 147: You can tell flak8 to ignore a function:

```
# flake8: noqa: C901
def func(): pass
```

I would revert your refactor and add the policy stuff to the old code, and keep all this logic in one function.

tests/barriers/test_history.py
```diff
@@ -35,7 +35,7 @@
 
 
 class TestBarrierHistory(APITestMixin, TestCase):
-    fixtures = ["barriers", "categories", "users"]
+    fixtures = ["barriers", "categories", "users", "policy_teams"]
```
abarolo, line 38: 👍

api/barriers/models.py
EPedley: Thanks I'll do that! How do you know when is a good time to ignore Flake 8?

osimuka (approved):

abarolo (approved):

ferozerub (approved):

PR #900, opened 2024-09-17
The last of us 2

abarolo:


osimuka (approved):

api/related_barriers/constants.py
```diff
@@ -1,7 +1,7 @@
 from collections import namedtuple
 
 SIMILARITY_THRESHOLD: float = 0.19
-SIMILAR_BARRIERS_LIMIT: int = 5
+SIMILAR_BARRIERS_LIMIT: int = 5000
```
ferozerub, line 4: Why is this so high?

api/related_barriers/serializers.py
```diff
@@ -3,13 +3,22 @@
...
+    summary = serializers.CharField()
+    title = serializers.CharField()
+    barrier_id = serializers.CharField()
+    reported_on = serializers.DateTimeField()
+    modified_on = serializers.DateTimeField()
+    status = StatusField()
+    location = serializers.SerializerMethodField()
+    similarity = serializers.FloatField()
+
+    def get_location(self, obj):
```
ferozerub, line 16: Not sure what we are doing here? Perhaps we should consider location_display field and status_display fields here to save lookups in the front-end

ferozerub (approved):

PR #913, opened 2024-10-07
Tss 2150 next steps dashboard iteration

Santino-Trade:


api/user/views.py
```diff
@@ -166,6 +166,11 @@ class DashboardUserTasksView(generics.ListAPIView):
     third_friday_date = None
     first_of_month_date = None
 
+    # Set variables to track a users relation to barriers in the loop
+    recipient_is_owner = False
```
abarolo, line 170: I dont think you should store state of individual users in the view.

Could perhaps return these as a tuple from the function `def set_user_barrier_relations(..) -> tuple`

abarolo: `is_owner, is_approver, is_publisher` = self.get_user_barrier_relations(recipient, barrier)`

Or you can put the function on the Barrier model itself?

api/user/views.py
```diff
@@ -271,6 +280,28 @@ def set_date_of_third_friday(self):
...
+        barrier_relations = {
+            "recipient_is_owner": False,
+            "recipient_is_approver": False,
+            "recipient_is_publisher": False,
+        }
+        for user in barrier_team_members:
+            if user.user_id == recipient.id and user.role == "Owner":
+                barrier_relations["recipient_is_owner"] = True
+
+        recipient_groups = recipient.groups.all()
```
abarolo, line 296: This gets called multiple times even though you only need to do it once right? Maybe pass it in as an argument

api/user/views.py
```diff
@@ -189,26 +200,24 @@ def get(self, request, *args, **kwargs):
             # Check if the barrier is overdue for publishing
             self.check_publishing_overdue(barrier)
 
+            # Establish the relationship between the recipient and barrier
+            barrier_relations = self.set_user_barrier_relations(recipient, barrier)
```
abarolo, line 204: can you rename recipient to user?  `user = request.user` especially since below you call the queryset `users_barriers`, so you have 2 different names for the same thing.

Also imho it is probably better to return as tuple, using a dict just hides the response schema and makes the response ambiguous. It also makes the conditions cleaner as below:
```
is_owner, is_approver = func()

if is_owner:
   pass
if is_approver:
   pass
```

abarolo (approved):

osimuka (approved):

PR #915, opened 2024-10-09
Adding sso_user_id to return in the user serializer

Santino-Trade:


osimuka (approved):

api/user/serializers.py
```diff
@@ -124,6 +126,12 @@ def get_permissions(self, obj):
             .values_list("codename", flat=True)
         )
 
+    def get_sso_user_id(self, obj):
+        if obj.profile.sso_user_id:
+            return obj.profile.sso_user_id
+        else:
```
abarolo, line 132: You can remove the else condition here. `return None` is the default behaviour of python functions

abarolo: You could even just do:

```
def get_sso_user_id(self, obj):
    return obj.profile.sso_user_id
```

abarolo: Or even:

`sso_user_id = serializers.CharField(source="profile.sso_user_id", null=True)`

Maybe add it to one of the existing tests too

abarolo (approved):

PR #923, opened 2024-10-21
tss-2202- Add fields to user profile

ferozerub:
Add Sectors, Organisations and Policy teams to user profile


tests/user/test_serializers.py
```diff
@@ -66,3 +66,28 @@ def test_group_modifications_logged(self, patched_logger):
...
+                    "trading_blocs": ["TB00003"],  # Asia pacific
+                    "overseas_regions": [
+                        "04a7cff0-03dd-4677-aa3c-12dd8426f0d7"
+                    ],  # Asia Pacific
+                    "policy_teams": [1, 4],
+                }
+            },
+        )
+        assert serializer.is_valid()
+        serializer.save()
```
osimuka, line 93: Do you want to check the data after save
`serializer.data` to see if the fields align?

osimuka (approved):

api/user/serializers.py
EPedley, line 1: By removing the if statements in the profile_update method we can send an empty list from the Frontend which means users can remove all policy teams.

api/user/serializers.py
```diff
@@ -164,8 +173,40 @@ def get_permissions(self, obj):
...
+            # update the related profile
+            sectors = profile_update.pop("sectors", None)
+            policy_teams = profile_update.pop("policy_teams", None)
+            organisations = profile_update.pop("organisations", None)
+            countries = profile_update.pop("countries", None)
+            trading_blocs = profile_update.pop("trading_blocs", None)
+            overseas_regions = profile_update.pop("overseas_regions", None)
+            # get current profile
+            profile = instance.profile
+            if sectors:
```
EPedley, line 190: If we remove the if statements here we can send an empty list from the frontend which removes all policy teams.

api/user/serializers.py
```diff
@@ -164,8 +173,43 @@ def get_permissions(self, obj):
...
+            trading_blocs = profile_update.pop("trading_blocs", None)
+            overseas_regions = profile_update.pop("overseas_regions", None)
+
+            # get current profile
+            profile = instance.profile
+            if sectors is not None:
+                profile.sectors = sectors
+            if policy_teams is not None:
+                profile.policy_teams.clear()
+                if policy_teams:
```
Santino-Trade, line 195: The ifs and fors are causing the complexity format error - do we need 195 and 200 if we've already decided the value is not None? This might take us under the limit if we can do without

osimuka (approved):

PR #924, opened 2024-10-22
Adding API level permission controls to only allow admins to change c…

Santino-Trade:
…ertain user details


api/core/permissions.py
```diff
@@ -28,3 +34,41 @@ def has_object_permission(self, request, view, obj):
...
+            # the user is an Admin, or return permission denied.
+            if field in ADMIN_PROTECTED_USER_FIELDS:
+                is_admin = False
+                for group in request.user.groups.all():
+                    if (
+                        group.name == "Administrator"
+                        or group.name == "Role administrator"
+                    ):
+                        is_admin = True
+                if is_admin:
```
Santino-Trade, line 61: condense line

api/user/constants.py
```diff
@@ -11,6 +11,16 @@ class UserRoles:
     ADMIN = "Administrator"
 
 
+ADMIN_PROTECTED_USER_FIELDS = [
+    "groups",
+    "username",
+    "password" "last_login",
```
Santino-Trade, line 17: comma

ferozerub (approved):

api/core/permissions.py
abarolo:
```
if field in ADMIN_PROTECTED_USER_FIELDS:
    return request.user.groups.filter(name__in=["Administrator", "Role administrator"]).exists()
```

api/core/permissions.py
```diff
@@ -28,3 +34,41 @@ def has_object_permission(self, request, view, obj):
...
+                    return True
+                else:
+                    return False
+
+        # At this point, we either have no protected fields being changed or are an
+        # admin and have already allowed progress. We can now check if the user is the
+        # owner of the record and is allowed to change the fields in the request.
+        if request.user.username == obj.username:
```
abarolo, line 69: You can just do:

```
return request.user.username == obj.username
```

osimuka (approved):

tests/user/test_views.py
```diff
@@ -277,6 +278,66 @@ def test_user_edit_add_new_profile(self):
...
+        edit_protected_response = api_client.patch(
+            url,
+            format="json",
+            data={"groups": [{"id": "4"}]},
+        )
+
+        assert edit_protected_response.status_code == 403
+
+        post_test_groups = update_user_test.groups.all()
+        assert len(post_test_groups) == 0
```
abarolo, line 334: You can removed everything below this line as the loop won't be executed below because the list is empty

api/core/permissions.py
```diff
@@ -28,3 +34,30 @@ def has_object_permission(self, request, view, obj):
...
+    """
+
+    def has_object_permission(self, request, view, obj):
+        # Read permissions are allowed to any request
+        if request.method in SAFE_METHODS:
+            return True
+
+        for field in request.data:
+            # Loop through fields in given data, if it is a sensitive field, check
+            # the user is an Admin, or return permission denied.
```
abarolo, line 52: One line of comment should be enough here: `# Check if admin`

api/core/permissions.py
```diff
@@ -28,3 +34,30 @@ def has_object_permission(self, request, view, obj):
...
+            if field in ADMIN_PROTECTED_USER_FIELDS:
+                return request.user.groups.filter(
+                    name__in=["Administrator", "Role administrator"]
+                ).exists()
+
+        # At this point, we either have no protected fields being changed or are an
```
abarolo, line 58: Same here, just `# Check if owner` or something

tests/user/test_views.py
abarolo: But probably better to do `assert update_user_test.groups.filter(name="Administrator").count() == 0` as another group would cause the count to go up

Santino-Trade: The test is specifically adding one group. There wont be others in there.

api/core/permissions.py
Santino-Trade: The following bit does more than just check for admin. The comment is accurate.

abarolo: `# Check if admin for sensitive groups`

tests/user/test_views.py
abarolo: You're right, but you are checking for 403 and that there is no admin group

PR #927, opened 2024-10-25
tss-2204: Barrier downloads filter params in body

abarolo:


api/barrier_downloads/views.py
```diff
@@ -15,6 +15,16 @@
 from api.barriers.models import Barrier, BarrierFilterSet
 
 
+
+class BarrierDownloadFilterBackend(DjangoFilterBackend):
+    def get_filterset_kwargs(self, request, queryset, view):
+        return {
+            'data': request.data.get("filters", {}),
```
Santino-Trade, line 22: Done some poking, looks like all we need here is:
'data': request.data,
As the filters are already passed in - also frontend just needs a slight change and it works fine. Locally. If the request data is empty it still passes through.

abarolo: Are you sure?  The original implementaiton does: `'data': request.query_params,`

The `filters` is passed from frontend is not the query params?

api/barrier_downloads/views.py
```diff
@@ -15,6 +15,15 @@
 from api.barriers.models import Barrier, BarrierFilterSet
 
 
+class BarrierDownloadFilterBackend(DjangoFilterBackend):
+    def get_filterset_kwargs(self, request, queryset, view):
```
Santino-Trade, line 19: I think we should have a comment here to point out why we're overriding the default behaviour, seems like the kind of thing that would be forgotten in a couple of years

Santino-Trade (approved):

osimuka (approved):

PR #987, opened 2025-04-23
Fix sort by relevancy and add relevancy score to search result panels

Santino-Trade:


api/barriers/models.py
```diff
@@ -1725,17 +1725,27 @@ def vector_search(self, queryset, name, value):
         text_search_ids = [str(b) for b in text_search_qs.values_list("id", flat=True)]
         barrier_scores = [(k, v) for k, v in barrier_scores if k not in text_search_ids]
 
-        return (
-            (text_search_qs | related_barrier_qs)
-            .annotate(
+        if queryset[0].ordering_value == "query_calculated":
```
abarolo, line 1728: Can you remove the duplicated query stuff? ie:

```
qs = (text_search_qs | related_barrier_qs).annotate(.....)

if queryset[0].ordering_valye == "":
    qs = qs.order_by("similarirty")
 
 return qs
```

ferozerub (approved):

PR #990, opened 2025-05-02
TSS 2669 - remove low scoring results from dashboard search

ferozerub:


api/barriers/models.py
```diff
@@ -1696,12 +1696,13 @@ def vector_search(self, queryset, name, value):
         """
 
         from api.related_barriers import manager as handler_manager
-        from api.related_barriers.constants import (
+        from api.related_barriers.constants import (  # SIMILARITY_THRESHOLD,
```
osimuka, line 1699: @ferozerub do you want to remove this import altogether since its commented out?

osimuka (approved):

api/barriers/models.py
ferozerub: Good spot

abarolo (approved):

